PR #35, opened 2014-06-07
Fixes Github TOC navigation

cirocosta:
With sectnum we can't navigate through the TOC on github. Would this affect pypi?


eliben:
The github/markup is fixed now.

PR #64, opened 2014-12-19
Update array dimension grammar

necase:
These changes update the pycparser grammar to align with C grammar to now allow parsing of constructs like
int f[restrict][5];
which is an example in the draft standard I consulted, and some other valid constructs as well.
I think these constructs are C99, not C11, as I consulted a draft standard numbered 1256 which didn't seem to mention C11 additions like _Generic.

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -983,41 +983,51 @@ class CParser(PLYParser):
         p[0] = p[2]
 
     def p_direct_declarator_3(self, p):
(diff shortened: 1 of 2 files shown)


necase:
After some unexpected time away, I have addressed your comments - the commit message now explains the change in a more obvious way than reading the diff does (your question was about where the handling of the static keyword went), and there is now a comment explaining why we can't use the slicing notation for the PLY object.

PR #69, opened 2015-02-17
adds visit_FuncDecl method to CGenerator

ssfrr:


eliben:
Is there a test that can trigger this?

ssfrr:
Just pushed a commit with some testing. I feel like there's a simpler way to exercise this functionality, but I'm not super familiar with the codebase so building a FuncDecl node from scratch is a little over my head, so I'm parsing it from code.

For context, the way I'm using pycparser is to parse header files and generate stubs, so I wanted a way to give a FuncDecl node to a CGenerator instance instead of giving it a full AST. In general being able to generate code for subtrees instead of the full AST seems really useful, I'm not sure if people are using it that way or not.

eliben:
Please squash into a single commit and I'll merge this

ssfrr:
squashed. should be good to go.

eliben:
Thanks for the contribution!

PR #74, opened 2015-04-19
Allow binary constants (e.g.: 0b01010)

Konstanty:
Here is a small patch to allow binary constants (like the existing octal and hex constants)


Konstanty:
Thanks for the review.

I've put the tag on the same line, and committed the parser test as well.
I'm not sure how github works with a "forced-update" version of the branch - does it require a new pull request, or is it automatically updated? (3bdbfdc32d6d9e97b867527f7e5be6a712d1edb6 is the new hash)

PR #78, opened 2015-04-22
Add a functional test

ianw:
Although consumers of pycparser should probably check they don't break
with new versions before relying on them, it would be nice if
pycparser also did a little sanity checking of arguably it's main
consumer; the cryptography->cffi->pycparser dependency chain.

This does a quick install and runs the cryptography tests; enough to
ensure there's no huge breakage.

(travis fails on the 3.2 build as idna isn't compatible with Python
3.2 due to it using u"foo")

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,7 @@ python:
   - "3.2"
   - "3.3"
   - "3.4"
-script: python tests/all_tests.py
 
+script:
+  - python tests/all_tests.py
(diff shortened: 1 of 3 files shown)


eliben:
mmm... I'll have to think about this :)

public:
I think we (i.e. @pyca) are happy enough to push bug reports upstream. We can always pin a version (it wouldn't be the first time) if a new pycparser release broke our integration.

PR #79, opened 2015-04-28
RFC: Allow empty declarations.

Konstanty:
I've been needing to parse empty declarations, however I am not sure of the correct way to implement this. The code in this pull request allows it to be parsed, but the output may not be 'correct'.

This is for statements like:
- int abc[3] = {}

Thanks in advance for the review.


eliben:
When does this need arise? Are empty initializer lists allowed by the standard?

Konstanty:
You are right - it is not part of C99.
It seems to be an extension which both clang and gcc implement.

eliben:
Fixed in c830da8

PR #84, opened 2015-06-09
Added taboutputdir parameter to control outputdir for tab files

shai-xio:
We are using a pycparser-derived parser in our  build process, and without this argument, it tended to leave {lex,yacc}tab.py files lying around. This way we can make sure they are generated as needed and cleaned up properly.


eliben:
Looks good, thanks

PR #86, opened 2015-06-11
Eliminate references to undefined initializer_list_opt

inducer:
The symbol `initializer_list_opt` is not defined anywhere. Mysteriously, pycparser's tests pass nonetheless, but pycparserext breaks.


inducer:
Never mind, sorry for the noise.

eliben:
Yeah, I realize this can be a bit confusing. Lately I'm not a big fan of such magic techniques which make the code un-greppable in some cases. On the other hand, it saves a fair amount of boilerplate code. It's a tradeoff.

PR #92, opened 2015-07-08
Modify AST calculation to include column values.

bbej84:
Column values were not being set in the AST.  Changes to pycparser/c_parser.py now set the coord.column value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ parser.out
 .tox
 utils/z.c
 *.egg-info
+*.swp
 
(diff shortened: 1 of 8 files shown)


eliben:
It seems that this pull request breaks the tests. Also, it doesn't seem to have any tests of its own. Finally, please rebase to squash into a single commit. All merge requests should have a single commit in them.

bbej84:
I fixed the code to pass the unit tests and added some tests to test_c_parser.py.  Let me know if more tests are required.

eliben:
Tests are still failing, as you see. Also, the PR is still not squashed into a single commit. I will not accept multiple commits, sorry.

PR #95, opened 2015-08-29
Add support for C11 atomics.

esoma:
Adds the _Atomic type qualifier & specifier and atomic headers. I'll address a couple of design decisions that I made:

The first is the split between the ATOMIC_QUALIFIER and ATOMIC_SPECIFIER tokens, rather than just using a single _ATOMIC keyword token. According to the C11 spec: *"If  the _Atomic keyword is immediately followed by a left parenthesis, it is interpreted as a type specifier (with a type name), not as a type qualifier"* (6.7.2.4). It seemed easier to let the lexer handle this with a quick look-ahead. I did a cursory test with a single token, but it results in a shift/reduce conflict (I imagine there's a way to write the grammar around this, but that's not my forte). Tests still pass, but I think it's more explicit/robust this way.

In order to meet the condition that *"The type name in an atomic type specifier shall not refer to an array type, a function type, an atomic type, or a qualified type."* (6.7.2.4) the parser needed to become aware of the actual type represented by a typedef, so I modified the way the internal _scope_stack works. Rather than holding True/False it holds the c_ast.Typedef or None in the case of identifier objects. This allows the production for the atomic specifier to inspect the type to make sure it follows the rules.

I added *stdatomic.h* and *uchar.h*, but I kept the macros and typedefs in their respective files, despite convention to put them in the *_fake_x.h* files. I figured, at least in the case of the atomic types, there was potential for naming conflicts with libraries that are still in C99 land, so excluding those from always getting included could be a good thing. For the *ATOMIC_X_LOCK_FREE* macros I chose the value of 1 because *"a value of 1 indicates that the type is sometimes lock-free"* (7.17.5), which I think best reflects pycparser's usage in which a potential target compiler's support is unknown.

### DIFF
--- a/examples/cdecl.py
+++ b/examples/cdecl.py
@@ -84,6 +84,8 @@ def _explain_type(decl):
         if decl.dim: arr += '[%s]' % decl.dim.value
 
         return arr + " of " + _explain_type(decl.type)
+    elif typ == c_ast.Atomic:
+        return '_Atomic ' + _explain_type(decl.type)
 
     elif typ == c_ast.FuncDecl:
         if decl.args:
@@ -100,7 +102,7 @@ if __name__ == "__main__":
     if len(sys.argv) > 1:
         c_decl  = sys.argv[1]
     else:
-        c_decl = "char *(*(**foo[][8])())[];"
+        c_decl = "char *(*(**foo[][8])(_Atomic(int*)))[];"
 
     print("Explaining the declaration: " + c_decl + "\n")
     print(explain_c_declaration(c_decl) + "\n")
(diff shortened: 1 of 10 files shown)


eliben:
Thanks, I'll take a look at it in a few days

eliben:
I have two big-picture issues here, before I dive into the detailed code review:

1. I don't see why this should be handled in the lexer and not the parser. Doesn't the standard provide the exact grammar rules for atomic, which can then be incorporated into the parser? Note that atomic is making `t_ID` 2-3 times longer, which is a shame. Semantic inference should be done in the parser, unless there's a strong reason (context sensitivity) to do it otherwise.

2. pycparser doesn't do type checking and other validation of the AST. It will let you redefine types, variables and so on. So I don't think the extensive validity checking of Atomic, which requires you to modify the scope stack, is necessary. It significantly complicates the code. pycparser assumes it sees valid C code, and validation can be added as an AST pass after the AST is built, if needed.

PR #100, opened 2015-10-01
Support universal wheels

graingert:


kevin-brown:
It's awesome to see that this has been enabled, as there shouldn't be anything major holding this project back, but it appears as though no wheels have been pushed up for pycparser to PyPI.

eliben:
Pardon my ignorance, but I have never pushed wheels to PyPI before. How do I do this?

graingert:
python3 setup.py sdist bdist_wheel upload

On 1 Oct 2016 22:55, "Eli Bendersky" <notifications@github.com> wrote:

Pardon my ignorance, but I have never pushed wheels to PyPI before. How do
I do this?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/eliben/pycparser/pull/100#issuecomment-250939988>, or mute
the thread

eliben:
OK, I think I uploaded a wheel. Does it make life easier for anyone? [pycparser is a pure-Python package]

kmwenja:
@eliben you kinda baited fate there. Thanks for the quick fix on #147

eliben:
I don't know why pycparser *needs* binary wheels. And sorry about the breakage - sometimes I forget that transitively lots of stuff depends on pycparser.

kevin-brown:
Given that the universal wheels were a pretty solid failure, it looks like there is more to this than I thought.

> I don't know why pycparser needs binary wheels.

It doesn't _need_ binary wheels, they're just useful to have as otherwise they get generated on every build. http://pythonwheels.com/

----

There was a concern noted in #148 that I do want to echo

> It looks like the wheel just uploaded to PyPI was cut from a different commit than what the tar.gz was cut from.

This issue appeared to affect Python 2.7 only (based on looking through tracebacks in the _many_ linked tickets), which sounds to me like either a universal wheel can't be used (in which case, this PR should be reverted), or there are issues in whatever commit was used to generate the wheels that affects Python 2.7.

graingert:
Maybe try a wheel of the correct version?

graingert:
But upload it to the testing PyPI

rodrigc:
@eliben thanks for fixing this.  There is nothing wrong with having a pycparser wheel,
you just need to make sure you have your checkout from the correct tag when running the commands.

Is there new stuff in the top of tree of pycparser that causes stuff to break during install?

PR #101, opened 2015-10-12
added support for Pragma

cod3monk:
Is there any chance to include (optional) support for #pragma in the AST? I know it is not strictly C99, since it is supposed to be handled before and not through the AST.

Dirty patch included, just as an example.

### DIFF
--- a/pycparser/c_ast.py
+++ b/pycparser/c_ast.py
@@ -795,3 +795,15 @@ class While(Node):
 
     attr_names = ()
 
+class Pragma(Node):
+    __slots__ = ('string', 'coord', '__weakref__')
+    def __init__(self, string, coord=None):
+        self.string = string
+        self.coord = coord
+
+    def children(self):
+        nodelist = []
+        return tuple(nodelist)
+
+    attr_names = ('string', )
+
--- a/pycparser/c_lexer.py
+++ b/pycparser/c_lexer.py
@@ -171,7 +171,9 @@ class CLexer(object):
         'ELLIPSIS',
 
         # pre-processor
-        'PPHASH',      # '#'
+        'PPHASH',       # '#'
+        'PPPRAGMA',     # 'pragma'
+        'PPPRAGMASTR',
     )
 
     ##
@@ -274,7 +276,6 @@ class CLexer(object):
 
     def t_ppline_NEWLINE(self, t):
         r'\n'
-
         if self.pp_line is None:
             self._error('line number missing in #line', t)
         else:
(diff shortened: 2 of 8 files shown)


eliben:
Yes, I'm fine with including pragmas in the AST, as long as the change is conservative - i.e. doesn't break other code and if we encounter pragmas we can tokenize but not parse yet it shouldn't fail (similarly to the way it's done now).

cod3monk:
I added and modified the unit test accordingly. Any other requests? 
Do you think the additional token will break any code?

pycparser/c_ast.py
```diff
@@ -795,3 +795,15 @@ def children(self):
 
     attr_names = ()
 
+class Pragma(Node):
```
eliben: You are not supposed to modify this file directly. It's auto generated

pycparser/c_lexer.py
```diff
@@ -171,7 +171,9 @@ def _make_tok_location(self, token):
         'ELLIPSIS',
 
         # pre-processor
-        'PPHASH',      # '#'
+        'PPHASH',       # '#'
+        'PPPRAGMA',     # '#pragma'
```
eliben: PPPRAGMA is for 'pragma', not '#pragma', right?

pycparser/c_lexer.py
```diff
@@ -304,15 +305,18 @@ def t_pppragma_NEWLINE(self, t):
...
+    t_pppragma_ignore = ' \t'
 
+    def t_pppragma_STR(self, t):
+        '.+'
+        t.type = 'PPPRAGMASTR'
+        return t
+    
     @TOKEN(identifier)
-    def t_pppragma_ID(self, t): pass
+    def t_pppragma_ID(self, t):
```
eliben: Why is this still ignored?

pycparser/c_ast.py
cod3monk: Yes, I forgot to add the changed _c_ast.cfg file to the commit. Added in 95ee04fffc2530a7f9966b517d5d9173e91e5b0e

pycparser/c_parser.py
```diff
@@ -547,9 +547,17 @@ def p_external_declaration_4(self, p):
 
     def p_pp_directive(self, p):
         """ pp_directive  : PPHASH
+                          | PPPRAGMA
```
eliben: I'd have a separate parsing rule for pragma now that it's supported, leaving the other one as it was

tests/test_c_lexer.py
```diff
@@ -309,28 +309,52 @@ def test_preprocessor_line_funny(self):
...
+        #\tpragma {pack: 2, smack: 3}
         #pragma <includeme.h> "nowit.h"
         #pragma "string"
         #pragma somestring="some_other_string"
         #pragma id 124124 and numbers 0235495
         59
         '''
-
-        # Check that pragmas are ignored but the line number advances
+        # Check that pragmas are tokanized, including trailing string
```
eliben: s/tokanized/tokenized/

pycparser/c_lexer.py
cod3monk: Is unused and can safely be removed. Should I remove it?

eliben: Yes, go ahead

cod3monk:
Anything else missing?

eliben:
Looks good. Please squash into a single commit and put as much info as possible into the commit message; then it's good to go.

eliben:
Thanks!

Issue #102, opened 2015-10-29
Ignoring comments


cod3monk:
Would you accept a pull request which leads to ignoring comments rather than failing on them?

eliben:
What would be the goal of that? Macros would fail anyway, right?

Perhaps as a separate pre-processing step?

cod3monk:
To allow users of my package to have comments within their code, without pycparser complaining about it. I would just ignore them on the parser side.
Of cause I could just filter them out before handing the code to pycparser, but I suppose other people would find it useful as well.

eliben:
My question was - you expect users to pass unpreprocessed code through to pycparser? What about macros and other preprocessor-y things ("\" line endings, etc.), what happens to these?

I'd prefer to maintain separate preprocessing from parsing for now. Therefore, I could see a separate something added to filter out C/C++ comments that could optionally be run before the parsing happens

PR #103, opened 2015-11-02
added function to remove comments and macros

cod3monk:
pycparser.clean(...) supports removal of comments and macro, single and
multline. It preserves line numbers by replacing removed lines with
empty ones, but it does also remove comments contained in strings.

As mentioned in #102.


eliben:
If I were to accept some basic pre-processor functionality into pycparser, I'd really prefer it to be more powerful than this. As it is now, I don't see the big advantage of this method over some ad hoc solution by users. Comment removal is very simplistic. Macro usages are not expanded out, etc.

Issue #107, opened 2015-12-12
No coord for Pragma Node


guivigon:
Hi 

when parsing the following code:

#pragma stml loop_schedule
  for(i=0;i<N;i++)
    v[N] = w[N]*w[N];

Within "visit_Pragma(self, node)" function of my visitor, if I print "node.coord" it shows "None". I've checked the pycparser code and it seems is not initializing properly the node coord. I'm very new to pycparser and not confident about modifying the code. Could someone please fix this issue?

Thanks.

Best.

eliben:
@cod3monk - care to take a look?

cod3monk:
Will have look at it tomorrow.

PR #108, opened 2015-12-14
Fix for #107 "No coord for Prgama Node"

cod3monk:
See #107

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -557,9 +557,9 @@ class CParser(PLYParser):
                                     | PPPRAGMA PPPRAGMASTR
         """
         if len(p) == 3:
-            p[0] = c_ast.Pragma(p[2])
+            p[0] = c_ast.Pragma(p[2], self._coord(p.lineno(2)))
         else:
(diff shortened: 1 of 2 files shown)


pycparser/c_parser.py
```diff
@@ -557,9 +557,9 @@ def p_pppragma_directive(self, p):
                                     | PPPRAGMA PPPRAGMASTR
         """
         if len(p) == 3:
-            p[0] = c_ast.Pragma(p[2])
+            p[0] = c_ast.Pragma(p[2], self._coord(p.lineno(2)))
         else:
-            p[0] = c_ast.Pragma("")
+            p[0] = c_ast.Pragma("", self._coord(p.lineno(2)))
```
eliben: This looks weird - are you sure it's lineno(2) and not lineno(1) in this case? Actually, maybe both cases...

[a test for this case (both with and without string) would be good too]

cod3monk:
You were right, I forgot about the without-string  version and-naturally-it did not work. Should be good now.

eliben:
Thanks!

PR #109, opened 2015-12-15
Update c_generator to add {} around nested NamedInitializers

jamie-pate:
I've monkey-patched this change in my project which depends on pycparser: http://github.com/jamie-pate/jstruct . It allows nested NamedInitializer lists by calling `_visit_expr()` instead of `visit()` inside `visit_NamedInitializer`.


eliben:
Thanks for the contribution!

PR #111, opened 2016-01-20
Problem: can't parse code from https://github.com/zeromq/malamute

asokoloski:
Solution: more fake headers and some typedefs from asm-generic/int-ll64.h

Thanks for making a useful project!


asokoloski:
Thanks :)

PR #114, opened 2016-02-25
fixed pathes to c-sources in examples

akhtyamovrr:
There have been some wrong pathes in examples like `filename = 'examples/c_files/year.c'`

### DIFF
--- a/examples/func_calls.py
+++ b/examples/func_calls.py
@@ -40,7 +40,7 @@ if __name__ == "__main__":
         filename = sys.argv[1]
         func = sys.argv[2]
     else:
-        filename = 'examples/c_files/hash.c'
+        filename = 'c_files/hash.c'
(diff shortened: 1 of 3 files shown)


eliben:
Can you clarify why you think the current code is wrong?

akhtyamovrr:
Because your *.py example files are in 'examples' folder. And if you try to run it, then path to c-sources is 'c_files/*.c'. Isn't it?

eliben:
Examples should be run from the root dir of pycparser - they make other assumptions for this (for example look how -I is passed for the fake headers dir).

akhtyamovrr:
Ok, thanx for clarification

eliben:
No problem. I pushed a README in the examples/ directory to clarify this. Thanks for the effort anyway :)

PR #119, opened 2016-04-02
Add "evalparam" method to parse a single def

eqvinox:
Very useful for tools that want to hunt down details after loading a file.

### DIFF
--- a/examples/eval.py
+++ b/examples/eval.py
@@ -1,0 +1,30 @@
+#-----------------------------------------------------------------
+# pycparser: ceval.py
+#
+# Interactive definition explorer - loads a source file,
+# then accepts type snippets on input which are parsed
+#
+# Takes source filename as command line argument
+#-----------------------------------------------------------------
+
+import sys
+sys.path.extend(['.', '..'])
+
+from pycparser import parse_file, c_parser
+import readline
+
+if __name__ == "__main__":
+    filename = sys.argv[1]
+
+    parser = c_parser.CParser()
+    ast = parse_file(use_cpp = True, filename = filename, parser = parser)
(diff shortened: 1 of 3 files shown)


eliben:
It's not clear from the PR what this is doing exactly. No tests / examples / real use cases outside modifying the lexer & parser code

eqvinox:
git cde1c4f with more explanation & example code

My own use case is a code generator that loads header files as input and has a separate spec file which contains rules like `mapping "struct foo" { bla bla };` where I use evalparam() to process the "struct foo" bit.

eliben:
So the idea is to parse a file and then parse additional lines in its context? Like if the file defines "typedef int FOO" you want to be able to successfully evaluate "FOO myinteger1"?

Why do you need a special token for this? All you need is the typedef mapping from CParser, no?

eqvinox:
Yes, though input isn't restricted to typedefs - it can be a C standard datatype, a struct, an union, and pointers as well.  The typedef mapping is indeed all I need, so another way to do it would be create a new parser and feed in the previous mapping (which would also need some changes).

The rationale for the special token is completely independent - I want to evaluate the input as a single definition. Parameter definition works best (for me, at least) since it doesn't accept initializers, and accepts unnamed parameters, so that's the choice there.

On a wider perspective, this is IMHO an important utility function to have to be able to do useful work with a parsed file.  Another useful function would probably be `evalexpr()` to be able to parse a single `expression_opt` (with knowledge of typedefs).

eliben:
I understand, thanks for explaining. However, I don't think adding the special token to core pycparser is a good idea. I would consider a PR that allows feeding a type mapping into CParser. Though it's possible today with "fake headers", I agree that a more programmatic API makes sense.

PR #120, opened 2016-04-04
Add "get_parameter_parser()" to create 'targeted' Parser

eqvinox:
rewritten version without special Lexer symbol; does more PLY processing runs instead

---

this returns another instance of CParser that is created with a copy of
the parser's current known symbols, and set to function parameter
context as grammar starting symbol.  It's parse function can be called
with a string describing a single parameter, as shown in eval.py:

$ python eval.py fifo.h

> struct fifo \* const \* x
> Decl: x, [], [], []
>   PtrDecl: []
>     PtrDecl: ['const']
>       TypeDecl: x, []
>         Struct: fifo
> int
> Typename: None, []
>   TypeDecl: None, []
>     IdentifierType: ['int']

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -24,7 +24,10 @@ class CParser(PLYParser):
             yacc_optimize=True,
             yacctab='pycparser.yacctab',
             yacc_debug=False,
-            taboutputdir=''):
+            yacc_errorlog=None,
+            taboutputdir='',
+            start='translation_unit_or_empty',
+            identifiers=dict()):
         """ Create a new CParser.
 
             Some arguments for controlling the debug/optimization
@@ -66,9 +69,20 @@ class CParser(PLYParser):
                 Generate a parser.out file that explains how yacc
                 built the parsing table from the grammar.
 
+            yacc_errorlog:
+                Logger used for debug/warning/error output from PLY.
+
             taboutputdir:
                 Set this parameter to control the location of generated
                 lextab and yacctab files.
+
+            start:
+                YACC start symbol in the C grammar; can be used to parse
+                particular snippets.
+
+            identifiers:
+                Dictionary of known names;  values indicate whether a name
+                refers to a typedef (True) or a variable/symbol (False).
         """
         self.clex = CLexer(
             error_func=self._lex_error_func,
@@ -104,11 +118,12 @@ class CParser(PLYParser):
 
         self.cparser = yacc.yacc(
             module=self,
-            start='translation_unit_or_empty',
+            start=start,
             debug=yacc_debug,
             optimize=yacc_optimize,
             tabmodule=yacctab,
-            outputdir=taboutputdir)
+            outputdir=taboutputdir,
+            errorlog=yacc_errorlog)
 
         # Stack of scopes for keeping track of symbols. _scope_stack[-1] is
         # the current (topmost) scope. Each scope is a dictionary that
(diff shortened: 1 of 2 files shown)


eliben:
Why is the setting to parameter scope necessary?

In any case, I think it would be a good idea to make this part of the change orthogonal to feeding known types in.

eqvinox:
`struct foo` is not a valid input in `translation_unit_or_empty` context; also the intent here is to explicitly disallow a function definition or something else.

On the orthogonality I'm not sure I fully understand so I'm guessing -- would it address your comment if the function was named `get_context_parser` and `start` is passed in by the user? (So, the change would not contain the `'parameter_declaration'` string)

pycparser/c_parser.py
```diff
@@ -22,9 +22,12 @@ def __init__(
             lex_optimize=True,
             lextab='pycparser.lextab',
             yacc_optimize=True,
-            yacctab='pycparser.yacctab',
+            yacctab=None,
```
eliben: The current yacctab name is "known" in other places in pycparser - this is a subtle point especially w.r.t. installations that won't need to regenerate it. I'd leave the default yacctab name as it was before

pycparser/c_parser.py
```diff
@@ -118,12 +139,16 @@ def __init__(
         # saw: int name;
         # If 'name' is not a key in _scope_stack[n] then 'name' was not defined
         # in this scope at all.
-        self._scope_stack = [dict()]
+
+        # typecheck input and deepcopy twice so we can reset to original
+        # identifiers in parse()
+        self._identifiers = dict([(str(key), bool(val)) for key, val in identifiers.items()])
```
eliben: why isn't this just a copy?

pycparser/c_parser.py
```diff
@@ -118,12 +139,16 @@ def __init__(
...
+        self._scope_stack = [dict(self._identifiers.items())]
 
         # Keeps track of the last token given to yacc (the lookahead token)
         self._last_yielded_token = None
 
-    def parse(self, text, filename='', debuglevel=0):
+    def parse(self, text, filename='', debuglevel=0, reset_identifiers = True):
```
eliben: I don't think adding the reset_identifiers param is really needed?

pycparser/c_parser.py
```diff
@@ -138,13 +163,30 @@ def parse(self, text, filename='', debuglevel=0):
...
+            be able to identify typedef'd types.  This function can be used
+            to get another parser instance that can be used to "continue"
+            parsing (e.g. snippets) with context of the loaded file.
+
+            The "start" keyword argument is particularly useful in this
+            context, specifically the "parameter_declaration", "expression"
+            or "constant_expression" states.
+        """
+        return CParser(
+                identifiers = self._scope_stack[0],
```
eliben: Nit: no spaces around `=` in named param assignment

eliben:
OK, understood. I left some specific comments in the diffs. Also please add a test.

pycparser/c_parser.py
eqvinox: I expect that some user will want to feed in their own created dict, so I think it's better to "fail fast" when they supply invalid data, rather than raise a cryptic exception somewhere later in the middle of parsing. It needs to be copied somewhere anyway to have the "child" parser not affect a "parent" parser, so the str() and bool() seems to me a good way to validate input at small extra cost.

pycparser/c_parser.py
eqvinox: Really depends on the use case... if it still were "parameter_context" specific as before, I'd agree, but for a generic interface I think it's useful to have the parameter. With reset_identifiers=False, it's possible to chain together multiple loads without using `get_context_parser()`. But when actually using `get_context_parser()` the reset_identifiers=True behaviour is much more likely to be useful.

eqvinox:
Test coming in a while (probably a few days), I'm currently occupied otherwise.

eliben:
Any updates on this?

eliben:
Closing for now, since original poster seems to be AWOL. I don't find the added functionality here particularly important for core pycparser

Issue #129, opened 2016-06-17
Parse Error on variable name that is a typedef in an outer scope, when it is declared together with another variable


csarn:
Sorry for the long title, but this is a weird circumstance.
This one parses correctly:
```c
typedef int foo;
void f(){
  int foo;
}
```
Note that "int foo" is declared on its own.

This one gives ParseError: test.c:3:10: before: foo
```c
typedef int foo;
void f(){
  int a, foo;
}
```
Just because there is another variable declared in the same statement.

FYI, I stumbled upon this when trying to parse python's own C source code (3.5.1) with pycparser.
This error happens in floatobject.c.

eliben:
This is an interesting one :)

pycparser has some workarounds for common cases of the problem described in http://eli.thegreenplace.net/2011/05/02/the-context-sensitivity-of-cs-grammar-revisited

I guess this one was left out. Contributions welcome to fix it - I'm currently over-swamped with other work

csarn:
Could you give any hints on where i might start looking? I'm not too familiar with pycparser's internals, but would like to investigate and hopefully fix this bug.

eliben:
Look at the functions the parser parameterizes the lexer with; in particular `type_lookup_func` and the usage of a "scope stack" to handle the case where an identifier is a type in one scope but and ID in another

eliben:
Closing this as duplicate of #118

PR #130, opened 2016-06-21
Fix issue #129: allow reuse of typedef-name as identifier in multi-id…

csarn:
…entifier declaration

Issue is reported here:
https://github.com/eliben/pycparser/issues/129
The fix adds 'TYPEID' to the possible values of direct_declarator.
This results in several shift/reduce conflicts, but the tests still
pass.
There is also a new testcase identifying the issue.

I'm not sure whether the shift/reduce conflicts have to be addressed,
but as i have no clue how to avoid them, and the tests all pass, i'm suggesting my fix :)

Here are the actual conflicts:

> Generating LALR tables
> WARNING: 7 shift/reduce conflicts
> WARNING: 57 reduce/reduce conflicts
> WARNING: reduce/reduce conflict in state 19 resolved using rule (direct_declarator -> TYPEID)
> WARNING: rejected rule (typedef_name -> TYPEID) in state 19
> WARNING: reduce/reduce conflict in state 83 resolved using rule (declarator -> pointer TYPEID)
> WARNING: rejected rule (direct_declarator -> TYPEID) in state 83

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -994,6 +994,7 @@ class CParser(PLYParser):
 
     def p_direct_declarator_1(self, p):
         """ direct_declarator   : ID
+                                | TYPEID
         """
         p[0] = c_ast.TypeDecl(
             declname=p[1],
--- a/tests/test_c_parser.py
+++ b/tests/test_c_parser.py
@@ -1797,6 +1797,19 @@ class TestCParser_typenames(TestCParser_base):
             '''
         self.assertRaises(ParseError, self.parse, s5)
 
+        # naming a variable after an outer-scope type should work even if
+        # it is not declared on its own line
+        s6 = r'''
+            typedef int TT;
+            void f(){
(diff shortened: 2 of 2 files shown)


eliben:
More conflicts is bad :) Currently pycparser only has one conflict, which is the well known ELSE shift/reduce conflict in the C grammars. There should be no others.

In particular, reduce/reduce conflicts are horrible since they mean two rules can be reduced on one input, and which rule is reduced is chosen more or less arbitrarily.

csarn:
Damn, so i just got lucky that the chosen rules are the correct ones... I'll see if i can fix it.

eliben:
Any updates on this?

csarn:
Not yet. I wanted this for a hobby project, but life happened and that project is on hold.
I might still get around to finishing it, but I don't know when.

I'll close the pull request. When/If I finish this bugfix, I'll post a new one against a more recent version of pycparser.

PR #141, opened 2016-08-16
report filename if error is "At end of input"

cod3monk:
Very simple fix to print filename in case "At end of input" is reported and no line number is given.

Current behavior:
```python
>>> from pycparser import CParser
>>> p = CParser()
>>> p.parse('(', filename='foo.c')
[...]
pycparser.plyparser.ParseError: : At end of input
```

New behavior:
```python
>>> p.parse('(', filename='foo.c')
[...]
pycparser.plyparser.ParseError: foo.c: At end of input
```


eliben:
Thanks!

PR #145, opened 2016-09-09
Fix eliben/pycparser#87 : offsetof() support is incomplete

Ksero:
I tried to follow the grammar of GCC's extension, as documented [here](https://gcc.gnu.org/onlinedocs/gcc-5.3.0/gcc/Offsetof.html).


eliben:
Thanks!

PR #155, opened 2016-11-10
using with...: in subprocess.Popen()

gryken2014:
To be (a little ) more robust with sub processes.


gryken2014:
It's python 3.x only ...

is this a problem ?

eliben:
Yes, it's a problem. `pycparser` should also work on Python 2.x

gryken2014:
OK.

PR #156, opened 2016-12-19
parse_text, preprocess_text added

rpoisel:
As the name suggests, these are two convenience functions that make unit-testing my code working with the resulting AST easier (at least for me). 

It would also be possible to make the ```parse_file()``` and ```preprocess_file()``` functions invoke ```parse_text()``` and ```preprocess_file()``` respectively. However, this would change the behavior of this library as the CPP would be forced to always read from stdin.

### DIFF
--- a/pycparser/__init__.py
+++ b/pycparser/__init__.py
@@ -8,10 +8,89 @@
 # License: BSD
 #-----------------------------------------------------------------
 __all__ = ['c_lexer', 'c_parser', 'c_ast']
-__version__ = '2.17'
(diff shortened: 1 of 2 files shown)


eliben:
Thanks for the contribution; I don't see a need for these functions in pycparser, however. There are many ways this can be done, and I want to leave it up to users. I'm trying to keep the API surface of pycparser as minimal as possible, without exposing too many utility functions

rpoisel:
That's understandable. So I will move this functionality to my modules.

PR #158, opened 2017-01-12
Add support for the __int128 type.

manueljacob:
This type is not part of the core C99 or C11 standards, but is mentioned
in both documents under "Common extensions".


eliben:
Thanks!

Issue #160, opened 2017-01-25
Please Update PLY Dependency to 3.9


riggsd:
I recently wrote a long bug report to the `sounddevice` team, who passed the buck to `CFFI`, whose developers implicated `pycparser` or `ply`. After lots of digging, I found a bug in PLY which prevented its use in a zipped-up source-less distributable as created by `py2app` or `pyinstaller`. After smugly filing a bug report with the PLY project, I checked out their upstream source so I could submit a PR only to find that... my bug had already been fixed!

It turns out that `pycparser` includes a *copy* of PLY 3.8 in your source tree, while PLY 3.9 is (currently) the latest release. 

Might I humbly request that you consider updating your packaged version of PLY to 3.9? I am sorry to say that `pycparser` is sufficiently over my head that I don't feel comfortable attempting a PR myself, so allow me to simply offer sincere THANKS!

See also:
https://github.com/spatialaudio/python-sounddevice/issues/68
https://groups.google.com/forum/#!topic/python-cffi/4JM22wGhVr4
https://github.com/dabeaz/ply/issues/108

riggsd:
PR submitted, and "it works for me"... I have unit tests failing with `master` branch, so I can't be 100% sure that PLY 3.9 doesn't bring any regressions with it. :-)

PR #161, opened 2017-01-25
Merge upstream PLY 3.9 into pycparser tree.

riggsd:
This updates PLY to the latest upstream code. Fixes Issue #160.


eliben:
Isn't the version you're copying here 3.10 rather than 3.9?

riggsd:
Ehrrm... Apologies, I thought that I'd merged the 3.9 tag, but I actually merged HEAD, which is one inconsistently-numbered commit ahead. The package-level `__init__.py` declares version 3.9 while the sub-modules each declare 3.10. 

I'll resubmit with a checkout of the 3.9 release tag.

Issue #67, opened 2014-12-29
Dump AST of a C file as a JSON string


user1020:
Thanks Eli for writing this module.  Wonder if it's possible to dump ast of a C file in a JSON form?
It would be easier to apply some customized instrument.

PR #163, opened 2017-01-30
dump and load as json

mypalmike:
In the examples folder, I implemented this feature:

https://github.com/eliben/pycparser/issues/67

And also its reverse.

The consistency of your ast node code made this quite straightforward, and I was able to implement it without modifying any core code. I abused __slots__ slightly in order to discover all constructor fields of each node class (empty child fields do not show up in children() calls, so they do not get serialized). This could be avoided by adding "=None" defaults to all node constructor arguments that represent child fields.

I've manually tested this on some of the example c files (I'm having preprocessor integration issues which make it difficult to test more thoroughly), but I should probably add some automated tests.

This is a really neat project by the way - I'm interested in using it as part of implementing a compiler for a very C-like language.

### DIFF
--- a/examples/c_json.py
+++ b/examples/c_json.py
@@ -1,0 +1,64 @@
+from __future__ import print_function
+
+import json
+import sys
+import re
+
+# This is not required if you've installed pycparser into
+# your site-packages/ with setup.py
+#
+sys.path.extend(['.', '..'])
+
+from pycparser import parse_file, c_ast, c_parser, c_generator
+
+RE_CHILD_ARRAY = re.compile('(.*)\[(.*)\]')
+
+class CJsonError(Exception):
+    pass
+
+def to_dict(node):
+    """ Recursively convert an ast into a python dictionary. """
+
+    klass = node.__class__
+    attr_names = klass.attr_names
+
+    result = {}
+    result['node_class'] = klass.__name__
+    for attr in klass.attr_names:
+        result[attr] = getattr(node, attr)
+
+    for child_name, child in node.children():
+        # child strings are either simple (e.g. 'value') or arrays (e.g. 'block_items[1]')
+        match = RE_CHILD_ARRAY.match(child_name)
+        if match:
+            array_name, array_index = match.groups()
+            array_index = int(array_index)
+            # arrays come in order, so we verify and append.
+            result[array_name] = result.get(array_name, [])
+            if array_index != len(result[array_name]):
+                raise CJsonError('Internal ast error. Array {} out of order. '
+                    'Expected index {}, got {}'.format(
+                    array_name, len(result[array_name]), array_index))
+            result[array_name].append(to_dict(child))
+        else:
+            result[child_name] = to_dict(child)
+
+    return result
+
+
+def to_json(node, **kwargs):
+    return json.dumps(to_dict(node), **kwargs)
+
+
+def file_to_json(filename, **kwargs):
+    ast = parse_file(filename, use_cpp=True)
+    return to_json(ast, **kwargs)
+
+
+#------------------------------------------------------------------------------
+if __name__ == "__main__":
+    if len(sys.argv) > 1:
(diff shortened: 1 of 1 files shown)


eliben:
Thanks! Can you elaborate a bit on the default `=None` problem you're talking about - what's missing, etc. It's not clear to me how `=None` would help here

mypalmike:
Sure. Note: It's only when reconstituting the ast from json that the issue arises: converting from ast to json is not a problem.

Well, it does start with the ast to json conversion code in some sense. This code walks the tree and converts each node from a python object into a dict. Traversal uses the children() method. The children() method for each class has code that skips over empty child fields. e.g.:

`class Decl(Node):
    def __init__(self, name, quals, storage, funcspec, type, init, bitsize, coord=None):
        ...
        self.bitsize = bitsize

    def children(self):
        ...
        if self.bitsize is not None: nodelist.append(("bitsize", self.bitsize))`

Parsing funky.c for instance, there's a Decl which has no bitsize children. The resulting json has no mention of bitsize. This seems reasonable imo.

However, when reconstituting the ast from json, the code has to call the node contructors with all fields. Since the json doesn't have a "bitsize" key in this case, the contructor fails (IIRC it crashes with something like "8 parameters needed, only 7 given").

The "hack" is to use the slots field to fill in any missing key, value pairs in the kwargs with None values. It's not a terrible hack really, just a use of the slots field that's not its normal purpose.

A couple approaches occurred to me. The first, which I implemented, was to use the slots field to discover all fields the constructor would need. A second, which I mentioned, would involve default value of None in the constructors. Another approach which now occurs to me would be to change my code so that it introspects the class during the conversion to json, emitting null values for empty child fields rather than omitting them in the json.

eliben:
Ah I see now, thanks for explaining.

The reason I'd prefer not to add `None` defaults to all constructor fields is that it would make AST creation more error prone, IMHO. Since AST nodes are created programmatially, it's not too cumbersome to require all params explicitly.

As you say, your `__slots__` approach isn't too bad. Stuff like serialization always requires some sort of reflection, and `__slots__` is a reasonable reflection mechanism in this case since all nodes define it.

An alternative is to use `dir()` on the AST nodes *when you dump them* to make sure you explicitly serialize all fields even if they're `None`, and then the deserialization path may be dumber. Would this make the code simpler?

mypalmike:
Yeah, I think serializing fields that are None would be better. I will change this.

Also, any preference on the naming of "_nodetype" (vs. "nodetype" or something else...)? The leading underscore makes it show up first in each node when sorted alphabetically, but I could be convinced to rename it.

mypalmike:
I believe this is good to go.

examples/c_json.py
```diff
@@ -0,0 +1,167 @@
+from __future__ import print_function
```
eliben: Mind adding a short blurb explaining what this example does and how to use it, at the top? Similar to other examples...

examples/c_json.py
```diff
@@ -0,0 +1,167 @@
...
+    Memoized to avoid highly repetitive string manipulation
+
+    """
+    non_child_attrs = set(klass.attr_names)
+    all_attrs = set([i for i in klass.__slots__ if not RE_INTERNAL_ATTR.match(i)])
+    return all_attrs - non_child_attrs
+
+
+def to_dict(node):
+    """ Recursively convert an ast into a dict representation """
```
eliben: Briefly describe what the dict representation is

examples/c_json.py
```diff
@@ -0,0 +1,167 @@
...
+        def __missing__(self, key):
+            ret = self[key] = fn(key)
+            return ret
+    return memodict().__getitem__
+
+
+@memodict
+def child_attrs_of(klass):
+    """
+    Given a Node class, get a list of child attrs.
```
eliben: a list or a set?

eliben:
Yes, almost ready. Just a few minor comments

Issue #82, opened 2015-05-23
Not possible to serialize the AST


pakt:
`pickle.dump` can't handle ast because the `ast` class uses `__slots__` but doesn't implement `__getstate__`. 

It'd be useful to add that functionality. Use case: parse a large repository once and store the ast for further processing. This way the source tree is parsed only once, saving time.

eliben:
Sounds good - feel free to send a pull request

eliben:
Any updates here?

chriscz:
For anyone actually trying to do this, in the meantime https://pypi.python.org/pypi/dill might be useful, I haven't tested whether it works, but there's a good chance it will.

PR #172, opened 2017-02-20
Add example of serializing AST for #82

CtheSky:
For issue #82, It's possible to serializing AST nodes now.  

From documentation:

> There are currently 5 different protocols which can be used for pickling. The higher the protocol used, the more recent the version of Python needed to read the pickle produced.

Using `__slots__ `without implementing `__getstate__` and `__setstate__` needs a protocol version >= 2. The default version is 3 for python 3.x and 1 for python 2.7. While version 2 is available since python 2.3.

So actually it's a solved problem and I added a simple example.

### DIFF
--- a/examples/func_calls.py
+++ b/examples/func_calls.py
@@ -1,5 +1,5 @@
 #-----------------------------------------------------------------
-# pycparser: func_defs.py
+# pycparser: func_calls.py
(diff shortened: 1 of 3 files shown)


eliben:
Thanks!

Issue #93, opened 2015-07-20
in examples:cdecl.py


ivand58:
cdecl.py doesn't recognize the structure declaration, for e.g.
<code>
cdecl.py "struct P {int x; int y} p;"
</code>
returns 
<code>
return quals + _explain_type(decl.type)

TypeError: cannot concatenate 'str' and 'NoneType' objects
</code>

eliben:
The cdecl sample wasn't designed to support inline struct declarations. Patches welcome

PR #174, opened 2017-02-23
Add support for expanding struct and typedef -- Issue 93

CtheSky:
Add optional arguments to expand `struct` and `typedef` for issue #93,  some examples here:
```python
s1 = 'typedef int Node; const Node* (*ar)[10];'
print(cdecl.explain_c_declaration(s1))
print(cdecl.explain_c_declaration(s1, expand_typedef=True))
# ar is a pointer to array[10] of pointer to const Node
# ar is a pointer to array[10] of pointer to const int

s2 = """
struct point {
  int x;
  int y;
};

struct rect {
  struct point pt1;
  struct point pt2;
} r;
"""
print(cdecl.explain_c_declaration(s2))
print(cdecl.explain_c_declaration(s2, expand_struct=True))

# r is a struct rect 
# r is a struct rect containing {pt1 is a struct point containing {x is a int, y is a int}, pt2 is a struct point containing {x is a int, y is a int}}

s3 = """
typedef struct {
  int x;
  int y;
} point;

struct rect {
  point pt1;
  point pt2;
} r;
"""
print(cdecl.explain_c_declaration(s3))
print(cdecl.explain_c_declaration(s3, expand_struct=True))
print(cdecl.explain_c_declaration(s3, expand_struct=True, expand_typedef=True))

# r is a struct rect 
# r is a struct rect containing {pt1 is a point, pt2 is a point}
# r is a struct rect containing {pt1 is a struct containing {x is a int, y is a int}, pt2 is a struct containing {x is a int, y is a int}}

s4 = 'struct P {int x; int y;} p;'
print(cdecl.explain_c_declaration(s4))
# p is a struct P containing {x is a int, y is a int}
```

### DIFF
--- a/examples/cdecl.py
+++ b/examples/cdecl.py
@@ -12,9 +12,23 @@
 #
 # For example:
 #
-# 'typedef int Node; const Node* (*ar)[10];'
-# =>
-# ar is a pointer to array[10] of pointer to const Node
+#   c_decl = 'typedef int Node; const Node* (*ar)[10];'
+#
+#   explain_c_declaration(c_decl)
+#   => ar is a pointer to array[10] of pointer to const Node
+#
+# struct and typedef are expanded when according arguments are set:
+#
+#   explain_c_declaration(c_decl, expand_typedef=True)
+#   => ar is a pointer to array[10] of pointer to const int
+#
+#   c_decl = 'struct P {int x; int y;} p;'
+#
+#   explain_c_declaration(c_decl)
+#   => p is a struct P
+#
+#   explain_c_declaration(c_decl, expand_struct=True)
+#   => p is a struct P containing {x is a int, y is a int}
 #
 # Eli Bendersky [http://eli.thegreenplace.net]
 # License: BSD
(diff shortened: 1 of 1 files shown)


examples/cdecl.py
```diff
@@ -95,6 +116,65 @@ def _explain_type(decl):
...
+                ('containing {%s}' % members if members else ''))
+
+
+def _expand_struct_typedef(decl, file_ast, expand_struct=False, expand_typedef=False):
+    """Recursively expand struct & typedef if according argument is true,
+       also check using undeclared struct & type and throw Exception
+    """
+    typ = type(decl)
+
+    if typ in (c_ast.Decl, c_ast.TypeDecl, c_ast.PtrDecl, c_ast.ArrayDecl):
```
eliben: I'd prefer the expansion to return a new decl rather than modify the existing one in place. Do you think that would be much more involved?

CtheSky: I have made it return a new decl. 

And I wonder that if expansion is only used in `explain_c_declaration`, then users won't have access to that node. Does it means that saying it "much more involved"  is based on the fact that it should be public to users and thus having the original node unchanged after expansion becomes a flexibility?

Just to check whether I get your idea and thanks in advance.

PR #178, opened 2017-03-09
Add column support in c_parser

serpilliere:
Hi!

This PR adds the support of `coord.column` in the ast parser.

It's inspired from the #92, by @bbejot. The main difference is that it only
fixes the missing `column`: it doesn't not add the `end_lineno` nor the
`end_column`.

Note: As the `_coord` function is not really used any more (only two calls remaining),
tell me if you prefer removing it in favour of `._token_coord`.

### DIFF
--- a/pycparser/plyparser.py
+++ b/pycparser/plyparser.py
@@ -51,6 +51,17 @@ class PLYParser(object):
                 line=lineno,
                 column=column)
 
+    def _token_coord(self, p, token_idx):
+        """ Returns the coordinates for the YaccProduction objet 'p' indexed
+            with 'token_idx'. The coordinate includes the 'lineno' and
+            'column'. Both follow the lex semantic, starting from 1.
+        """
+        last_cr = p.lexer.lexer.lexdata.rfind('\n', 0, p.lexpos(token_idx))
+        if last_cr < 0:
+            last_cr = -1
+        column = (p.lexpos(token_idx) - (last_cr))
+        return self._coord(p.lineno(token_idx), column)
+
     def _parse_error(self, msg, coord):
         raise ParseError("%s: %s" % (coord, msg))
 
--- a/tests/test_c_parser.py
+++ b/tests/test_c_parser.py
@@ -88,10 +88,13 @@ class TestCParser_base(unittest.TestCase):
     def setUp(self):
         self.cparser = _c_parser
 
-    def assert_coord(self, node, line, file=None):
+    def assert_coord(self, node, line, column=None, file=None):
         self.assertEqual(node.coord.line, line)
+        if column is not None:
+            self.assertEqual(node.coord.column, column)
         if file:
             self.assertEqual(node.coord.file, file)
+
 
 
 class TestCParser_fundamentals(TestCParser_base):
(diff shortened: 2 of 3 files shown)


pycparser/plyparser.py
```diff
@@ -51,6 +51,13 @@ def _coord(self, lineno, column=None):
                 line=lineno,
                 column=column)
 
+    def _token_coord(self, p, token_idx):
```
eliben: this function could use a docstring, as well as comments inside

tests/test_c_parser.py
```diff
@@ -88,12 +88,15 @@ def parse(self, txt, filename=''):
     def setUp(self):
         self.cparser = _c_parser
 
-    def assert_coord(self, node, line, file=None):
+    def assert_coord(self, node, line, column, file=None):
         self.assertEqual(node.coord.line, line)
+        if column:
```
eliben: what if column == 0?

tests/test_c_parser.py
```diff
@@ -134,20 +137,20 @@ def test_empty_toplevel_decl(self):
...
         t1 = """
         int a;
         int b;\n\n
         int c;
         """
         f1 = self.parse(t1, filename='test.c')
-        self.assert_coord(f1.ext[0], 2, 'test.c')
-        self.assert_coord(f1.ext[1], 3, 'test.c')
-        self.assert_coord(f1.ext[2], 6, 'test.c')
+        self.assert_coord(f1.ext[0], 2, 13, 'test.c')
```
eliben: It's not very intuitive that this is column 13... How about wrapping parse to strip leading whitespace as long as the first line or something like that, to get a more expectedoffset?

pycparser/plyparser.py
serpilliere: You are right! I will add it.

tests/test_c_parser.py
serpilliere: In fact, the `column` implemented keeps the semantic of lex, so the column always starts at 1, not 0.
Do you want me to test against `None` (plus an assert against 0)?

By the way, do you agree to keep the semantic of lex (starting column AND lineno at 1) or do you want a modification to starts column at 0?

tests/test_c_parser.py
serpilliere: in fact, as the `column` starts at 1, there is no spaces counted in tokens. the 13 starts on the `a`:
```
  TypeDecl: a, []
    IdentifierType: ['int']
```

Again, if you don't agree with the lex column start, don't hesitate to tell me and I will modify the behavior.

serpilliere: I used an (behemoth.h from https://github.com/snare/ida-efiutils/blob/master/behemoth.h) input file of 21945 lines (preprocessed). cProfile For the c_parse.py:parse:

without patch: 1.774s/1.784s/1.801s
with patch: 1.847s/1.857s/1.891s

So about 4% of speed lost

As the function is often called in this case (about 31000), I suspect a little overhead of cprofile.

Raw value on the whole python script:
without patch: 1.997s/2.030s/2.093s
with patch: 2.040s/2.053s/2.090s

(Note the max time of both runs: maybe the timing measures are not really correct here)

tests/test_c_parser.py
eliben: Yes, a default value of None and a test against None should be better, I think

Starting column at 1 is fine

tests/test_c_parser.py
eliben: I meant all the leading space, not 0-1 starting. Without columns it wasn't bothersome but now it's hard to see why this should be 13 and not some close-to-that number because everything is indented. To keep this patch small, I'm OK with fixing this somehow in a follow-up

serpilliere:
- [X] add `is None` for column
- [X] comment for `_token_coord`

I also modified the regression tests in which the file name was not present to comply with the fixed `assert_coord` api.

eliben:
Thanks, looks good

PR #181, opened 2017-03-15
Add Mir typedefs in fake headers

realitix:


eliben:
I think this PR conflicts with the others you sent.

Can you just send one single PR adding all these headers? That would be OK with me and make merging easier

realitix:
Indeed, I'm going to group all pull requests today.

eliben:
Thanks. I usually prefer small PRs, but in this case there's no point. These changes are fairly trivial and automatically self-conflicting, so I'm OK pulling large changes to the fake headers at once

PR #183, opened 2017-03-16
Add fakedef

realitix:
Add mir and xcb typedef


realitix:
I'm using these headers to make a Vulkan wrapper.

PR #159, opened 2017-01-14
Add argument to CParser.__init__ for overriding the yacc start symbol.

manueljacob:
This has been very useful for me to be able to parse e.g. expressions directly.

PR #190, opened 2017-05-18
Add support for continuing parsing & different start symbol

eqvinox:
Updated version of #120, with tests & demo. Also provides a superset of the (reverted) #159.

All feedback from both #120 and #159 should be addressed.

### DIFF
--- a/examples/eval.py
+++ b/examples/eval.py
@@ -1,0 +1,32 @@
+#-----------------------------------------------------------------
+# pycparser: ceval.py
+#
+# Interactive definition explorer - loads a source file,
+# then accepts type snippets on input which are parsed
+#
+# Takes source filename as command line argument
+#-----------------------------------------------------------------
+
+import sys
+sys.path.extend(['.', '..'])
+
+from pycparser import parse_file, c_parser
+import readline
+
+if __name__ == "__main__":
+    filename = sys.argv[1]
+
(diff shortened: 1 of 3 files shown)


eliben:
Sorry, but I'm still not too excited about this kind of change. I feel it addresses a very specific use-case of pycparser which isn't very generally useful, but it adds considerable complexity in the trickiest part of the project - interaction with PLY and its Yacc table generation; this backfired many times in the past due to the huge stack of projects implicitly dependent on pycparser (because of cffi), and I simply don't have the bandwidth to deal with these issues again.

eqvinox:
This is why my original implementation in #119 used a special token to switch the parser state. It looks odd, but is actually far cleaner than messing with PLY. I can switch back to that.

I believe the usefulness is sufficiently evident from 2 people independently needing the feature, cooking up patches, and submitting PRs.

eliben:
I feel a bit bad about rejecting PRs and it's great to see the work folks are willing to put into pycparser, but if there's any chance for me to continue maintaining this project with my dwindling free time, I have to aggressively restrict its scope.

On Github it's easy to keep enhancements to a project in a fork. https://github.com/inducer/pycparserext would be the prime example for this

eqvinox:
... or maybe you could ask someone for help in maintaining, like maybe the cffi people.

PR #198, opened 2017-07-07
Don't fail if docstrings are disabled

dsanders11:
The attribute `__doc__` will return `None` if running CPython in `-OO` mode [as it discards docstrings](https://docs.python.org/3/using/cmdline.html#cmdoption-OO).

FYI, this broke builds for me because `cryptography` uses `cffi` which uses `pycparser` at install time, and always pulls the latest version regardless of any dependency pinning in `requirements.txt`. It would be nice to prioritize this getting in a release ASAP to fix this issue.

### DIFF
--- a/pycparser/plyparser.py
+++ b/pycparser/plyparser.py
@@ -109,7 +109,10 @@ def _create_param_rules(cls, func):
             func(self, p)
 
         # Substitute in the params for the grammar rule and function name
-        param_rule.__doc__ = func.__doc__.replace('xxx', xxx).replace('yyy', yyy)
+        try:
+            param_rule.__doc__ = func.__doc__.replace('xxx', xxx).replace('yyy', yyy)
+        except AttributeError:
+            pass  # Docstring may not exist
         param_rule.__name__ = func.__name__.replace('xxx', xxx)
 
         # Attach the new method to the class


dsanders11:
Was not able to work around the build issue because it's a tangled web from `cryptography` to `cffi` to `pycparser` at build time, no easy way to pin the version of `pycparser` or switch to a fork.

alxchk:
Same story =\

eliben:
What's mysterious to me is how pycparser can ever work without docstrings? Since this is what PLY uses to get its parsing rules from. Or doesn't pycparser actually run in these scenarios?

dsanders11:
@eliben, I don't know about the internals of pycparser unfortunately, just that with 2.17 everything worked fine with `-OO` mode and 2.18 is where it broke.

Again, my use case is `cryptography` which uses `cffi` which uses `pycparser`. It's at install time of `cryptography` that the failure happens, so turning off `-OO` mode during install and turning it back on after does work around the issue.

eliben:
I still think this change plasters over the real problem rather than fixing it. If I apply this PR locally and run pycparser's tests with `-OO`, as expected I get:

```
$ python -OO tests/all_tests.py 
ERROR: ./pycparser/c_lexer.py:277: No regular expression defined for rule 't_ppline_NEWLINE'
ERROR: ./pycparser/c_lexer.py:289: No regular expression defined for rule 't_ppline_PPLINE'
ERROR: ./pycparser/c_lexer.py:301: No regular expression defined for rule 't_pppragma_NEWLINE'
ERROR: ./pycparser/c_lexer.py:306: No regular expression defined for rule 't_pppragma_PPPRAGMA'
ERROR: ./pycparser/c_lexer.py:312: No regular expression defined for rule 't_pppragma_STR'
ERROR: ./pycparser/c_lexer.py:247: No regular expression defined for rule 't_PPHASH'
ERROR: ./pycparser/c_lexer.py:326: No regular expression defined for rule 't_NEWLINE'
  File "tests/all_tests.py", line 15, in <module>
    'test_c_generator',
  File "/usr/lib/python2.7/unittest/loader.py", line 130, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python2.7/unittest/loader.py", line 91, in loadTestsFromName
    module = __import__('.'.join(parts_copy))
  File "/home/eliben/eli/pycparser/tests/test_c_parser.py", line 18, in <module>
    yacctab='yacctab')
  File "./pycparser/c_parser.py", line 88, in __init__
    outputdir=taboutputdir)
  File "./pycparser/c_lexer.py", line 66, in build
    self.lexer = lex.lex(object=self, **kwargs)
  File "./pycparser/ply/lex.py", line 910, in lex
    raise SyntaxError("Can't build lexer")
SyntaxError: Can't build lexer

```

So what is CFFI doing exactly that triggers the first but not the last? 

Maybe it's just *importing* pycparser without actually calling into it? If that's the case the real fix would be to avoid the class decorator that requires docstrings?

dsanders11:
Unfortunately I don't think things are going to get more clear, only more convoluted. The specific problem shows up at *install* for `cryptography`, which was one of the problems. It's going off and downloading `cffi` itself at install time which pulls the latest `cffi` and avoids any version pinning (really not good for repeatable builds, as I unfortunately found).

Here's the full output I see for the install failing, maybe you'll be able to gleen more info from it:

```
  Running setup.py install for cryptography: started
    Running setup.py install for cryptography: finished with status 'error'
    Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-QjVNrh/cryptography/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-m9tt6q-record/install-record.txt --single-version-externally-managed --compile:
    warning: no previously-included files matching 'yacctab.*' found under directory 'tests'
    warning: no previously-included files matching 'lextab.*' found under directory 'tests'
    warning: no previously-included files matching 'yacctab.*' found under directory 'examples'
    warning: no previously-included files matching 'lextab.*' found under directory 'examples'
    zip_safe flag not set; analyzing archive contents...
    pycparser.ply.lex: module references __file__
    pycparser.ply.lex: module MAY be using inspect.getsourcefile
    pycparser.ply.ygen: module references __file__
    pycparser.ply.yacc: module references __file__
    pycparser.ply.yacc: module MAY be using inspect.getsourcefile
    pycparser.ply.yacc: module MAY be using inspect.stack
    
    Installed /tmp/pip-build-QjVNrh/cryptography/.eggs/pycparser-2.18-py2.7.egg
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-QjVNrh/cryptography/setup.py", line 321, in <module>
        **keywords_with_side_effects(sys.argv)
      File "/usr/lib/python2.7/distutils/core.py", line 111, in setup
        _setup_distribution = dist = klass(attrs)
      File "/usr/lib/python2.7/site-packages/setuptools/dist.py", line 321, in __init__
        _Distribution.__init__(self, attrs)
      File "/usr/lib/python2.7/distutils/dist.py", line 287, in __init__
        self.finalize_options()
      File "/usr/lib/python2.7/site-packages/setuptools/dist.py", line 390, in finalize_options
        ep.load()(self, ep.name, value)
      File "/usr/lib/python2.7/site-packages/cffi/setuptools_ext.py", line 188, in cffi_modules
        add_cffi_module(dist, cffi_module)
      File "/usr/lib/python2.7/site-packages/cffi/setuptools_ext.py", line 49, in add_cffi_module
        execfile(build_file_name, mod_vars)
      File "/usr/lib/python2.7/site-packages/cffi/setuptools_ext.py", line 25, in execfile
        exec(code, glob, glob)
      File "src/_cffi_src/build_openssl.py", line 82, in <module>
        extra_link_args=extra_link_args(compiler_type()),
      File "/tmp/pip-build-QjVNrh/cryptography/src/_cffi_src/utils.py", line 61, in build_ffi_for_binding
        extra_link_args=extra_link_args,
      File "/tmp/pip-build-QjVNrh/cryptography/src/_cffi_src/utils.py", line 69, in build_ffi
        ffi = FFI()
      File "/usr/lib/python2.7/site-packages/cffi/api.py", line 63, in __init__
        from . import cparser
      File "/usr/lib/python2.7/site-packages/cffi/cparser.py", line 7, in <module>
        import pycparser
      File "/tmp/pip-build-QjVNrh/cryptography/.eggs/pycparser-2.18-py2.7.egg/pycparser/__init__.py", line 14, in <module>
        from .c_parser import CParser
      File "/tmp/pip-build-QjVNrh/cryptography/.eggs/pycparser-2.18-py2.7.egg/pycparser/c_parser.py", line 20, in <module>
        class CParser(PLYParser):
      File "/tmp/pip-build-QjVNrh/cryptography/.eggs/pycparser-2.18-py2.7.egg/pycparser/plyparser.py", line 95, in template
        _create_param_rules(cls, method)
      File "/tmp/pip-build-QjVNrh/cryptography/.eggs/pycparser-2.18-py2.7.egg/pycparser/plyparser.py", line 112, in _create_param_rules
        param_rule.__doc__ = func.__doc__.replace('xxx', xxx).replace('yyy', yyy)
    AttributeError: 'NoneType' object has no attribute 'replace'
```

Before the release of `pycparser` 2.18 the build would succeed and there have been no problems in usage (also running `-OO`).

FYI for replicating it may be best to use `export PYTHONOPTIMIZE=2` as that's what I have and that's what keeps it using the flag during `pip` installs.

Does a large amount of `pycparser` rely on being able to read docstrings for internal behavior? If it was trivial to change that, that would seem like the best course of action, but I have a feeling it may be a more fundamental design choice.

The reason we use `-OO` is it cuts memory usage per web worker in production by 10%, which is significant savings for us.

alxchk:
Same story for me. Only project which has dependency on pycparser is cryptography, and I'm not really sure that it really used in my case.

eliben:
@natezb that added this code in #169

natezb:
Like Eli said, `pycparser` is using `PLY` to generate its parser, and `PLY` uses the docstrings to get its parsing rules, so it's confusing how it could ever have worked without docstrings. I just took quick look at `PLY`'s source, and it looks like it ignores all `p_`-functions that are missing docstrings.

My first thought (like Eli) is that `pycparser` is perhaps only being imported, and not used for parsing. In this case, the production rules would be ignored, which won't directly produce an error. However, the code you discovered _is_ run at import-time.

Now that I think about it... are you using `cffi`'s API mode, i.e. actually using a c compiler? If so, it's probably not using `pycparser`, but is just importing it.

I'm not sure which project is the better one to patch. Since the parser is likely silently broken without docstrings as it is, maybe the proposed patch isn't so bad, as long as we log the issue like `PLY` does.

alxchk:
Well, cryptography says that it's not them )

natezb:
Well arguably it's nobody's bug, since the packages weren't designed to be run/installed in `-OO` mode. However, if we agree that it's worthwhile to support it, I see two main options:

1. Modify `cffi` to support API mode without requiring the use of `pycparser`. This may or may not be a simple change. ABI mode would still require `pycparser`, and therefore docstring support. To me, this is the "right" solution, but may be overly complicated in practice.

2. Sweep this under the rug in `pycparser` as suggested above. This is arguably the "wrong" way to do it, since (to my understanding) `pycparser` is useless without docstrings. This change would only be to allow `cffi` to import `pycparser` and instantiate a parser (since `cffi` appears to do _at least_ this much), but the parser would not be functional.

dsanders11:
@natezb, good discussion. Another issue I'd point out is the timeliness of resolving the issue. It cropped up with the 2.18 release (so the last week). `cffi` hasn't had a release since March, so pursuing 1 likely means this wouldn't be fixed for a while since any changes will likely be large enough they'd be worked into the normal release cycle rather than a hotfix kind of thing.

As such `pycparser` is in the better position to fix the regression in a more timely manner, releasing a 2.19 keeping the breakage time from 2.18 to a minimum. As such I'd favor 2 for at least the immediate future (we can add a warning instead of just a `pass`) as it fixes the regression right away, and makes it less time sensitive.

Just my 2 cents.

eliben:
Re timeliness, I just want to say that it's not very easy to release pycparser *precisely* because of issues like this one. It's impossible to foresee all the weird ways pycparser's dependencies get used and installed, and very strange problems pop up constantly.

If you ask me, it's madness that CFFI was last released in March but pycparser gets pulled in the latest revision. Did CFFI's release test with pycparser 2.18? Of course not, it didn't exist. So why don't they pin pycparser to a version they actually tested with? This sort of forward compatibility is imprudent.

Sorry for the rant, I hope you can see this is frustrating from my side - this is just a hobby project I'm maintaining on slivers of free time.

dsanders11:
@eliben, I totally agree, and understand how it's frustrating. I was also very frustrated, imagine my surprise when my build suddenly broken despite all dependencies being pinned, because of `cffi` shenanigans.

I just thought it more prudent to jump ahead to `pycparser` for the timeliness issue I mentioned, in addition to the fact that even if `cffi` did pin, at some point they'd want to update to `pycparser` 2.17+ and we'd be back to trying to find a solution for `-OO` somewhere in the grey area.

Didn't mean to cause undue grief, I thought this PR might be a quick fix to the situation but it's clearly more complex. It can currently be worked around by disabling `-OO` mode for install and enabling it again for actual usage. Not ideal, and it leaves a lot of `.pyc` files around in addition to the `.pyo` that get created by `-OO`, but it's manageable if it keeps you et al from tearing your hair out.

eliben:
I'm OK with helping out from pycparser's side, just wanted to find a more "real" solution.

@natezb what do you think about not having the decorator munge the docstrings? This decorator would run if c_parser is imported even without instantiating. If we don't use a decorator, I think the import is safe but if the user instantiates c_parser (which it apparently does not because otherwise PLY would fail) they'll just get the exception (they'd get it from PLY anyway)

natezb:
@eliben I'll have to take a closer look at this tonight, but it's my understanding that `CParser` _is_ actually instantiated, just never used.

It's certainly possible do to away with the decorators. I'll have to think about this more, but here are some options:

1. Do the templating in `CParser`'s `__init__` method, as the `opt` rules are handled
2. Eliminate templating altogether, and just add all of the functions explicitly (though there will be many of them)

eliben:
Thanks, (1) sounds reasonable if it will work

natezb:
It does appear that `CParser` is getting instantiated, so (1) won't help us in the current scenario (it was silly of me to suggest it, in retrospect).

Here's my current understanding of the situation. When a `CParser` is instantiated, it calls `yacc.yacc()` to generate the parser. This will go through `CParser`'s docstrings, but _will not_ raise any exceptions. It then checks if there's an existing `yacctab.py`, loading and returning it if it exists.

What this means is that, once the tables have been generated, docstrings are no longer needed by PLY. Thankfully, these generated tables are provided by the `sdist`s on PyPI.

If we want an immediate fix, we must be able to both import `pycparser` and instantiate `CParser` without error, given preexisting tables. I now see these options:

1. Accept the patch from @dsanders11 , or a variation on it. This allows import to always work without docstrings. If a pregenerated table exists, the parser will work as normal. If not, I think this will fail upon parser generation, but I'm not exactly sure (it might just generate a useless parser).

2. Eliminate all templating (same as option 2 above).

Whether you generate the methods via templating or write them explicitly, they will have no docstrings in `-OO` mode, so their effects are equivalent. With that in mind, my vote is strongly in favor of (1), since it's only a simple, minor change and leaves us with much less code.

eliben:
I've pushed a fix that's a bit more strict than this PR by issuing a warning in this case (the warning may be seen during installation, but it's mostly for other problems like someone forgetting a docstring on a parameterized rule).

Please test that this addresses your issue

dsanders11:
@eliben, I'll give it a test. Unfortunately can't test it in the exact reproduce case since it's a PITA to try to change the version of `pycparser` being pulled without forking `cffi` and then forking `cryptography` due to the weird install time stuff.

dsanders11:
LGTM. 👍

eliben:
Is there any chance @dsanders11 or @dannosaur can test it in something approaching the real deal? Your setups appear to be fairly unique given the low number of complaints. In the past when things really broke I had hundreds of people complaining at the same time (since cffi and cryptography are fairly popular).

Thanks for the help btw - and thanks to @natezb for helping thinking this through.

dannosaur:
I can certainly try it out on my environments to see if the original issue I reported has since gone away. I operate an exact replica of our production environment that I can play with and break if I so desire, so I'll pull master in on there and see if the problem persists.

I guess, given the conversation above, the fact that cffi is picky over installing the latest version from PyPi that I need to do some manual mungery of the virtualenv post-install to override the version it's installing with the HEAD of master, but I'll report back when I've been through it, probably Monday now.

eliben:
@dannosaur thanks in advance!

I'd still encourage folks to open an issue for CFFI asking to pin pycparser to a revision they actually tested with when releasing. I don't think that letting pycparser's revision be floating and running far ahead of cffi makes much sense for them.

dannosaur:
Unfortunately master doesn't resolve the issue using -OO (in my environment, at least). The error, however, has changed...

```
  File "/srv/www/project/current/.venv/lib/python2.7/site-packages/cffi/cparser.py", line 44, in _get_parser
    _parser_cache = pycparser.CParser()
  File "/srv/www/project/releases/20170719184420/.venv/src/pycparser/pycparser/c_parser.py", line 117, in __init__
    outputdir=taboutputdir)
  File "/srv/www/project/releases/20170719184420/.venv/src/pycparser/pycparser/ply/yacc.py", line 3352, in yacc
    raise YaccError('Unable to build parser')
pycparser.ply.yacc.YaccError: Unable to build parser
```

natezb:
@dannosaur please see my discussion of this issue above. Your problem is almost certainly because you're lacking the pre-built parse table that would come with a release of `pycparser`. You may be able to do this by first installing `pycparser` while *not* using `-OO` mode (I may be wrong about this).

dannosaur:
Basically the way I installed this was to hijack the venv from one of our staging servers after a deployment had completed, manually uninstalled the pycparser version that our chef run installed, then installed it direct from master using a standard pip install command (presumably this runs without any form of optimization whatsoever).

So, if the missing parse tables are supposed to be built at install/setup time - they weren't. If there's a secondary step post-install of pycparser required to do something in order to get it to build these tables, then that complicates matters since I (and likely many others) will have to invent something to make this happen before restarting backend servers (in our case, the virtualenv is refreshed after every deployment..it's why you see date stamps in the file paths in my comment above). This feels like a hack though.

Is there any way to determine if Python is running in `-OO` mode, and if so, rather than allow it to fall over at some other random point, if docstrings are truely required for this package to work correctly/at all, raise some other sort of exception indicating this? I switched our production environments to using `-O` and they still work perfectly fine (as expected given the differences between the 2 modes) with 2.18. I'm just wondering at this point which is the lesser of 2 evils...tear hair out over trying to fix this to run in `-OO` mode, or simply raise a helpful exception?

I agree with @eliben though that this is something that CFFI need to sort at their end as well, they shouldn't be requiring "whatever is the latest version" without testing it. But even then, it's only a matter of time before they release a new version, test their new package against pycparser 2.18+ and release it, then eventually causing this issue with those that are running `-OO`. All they've done by doing this is expedite the process of trying to figure this out now, rather than later.

eliben:
I'm leaning against a new release given the issues discussed recently. This is too hard to test reliably, and manifests only in environments I don't use and frankly make little sense. You can always pin to a git revision, if really needed.

I think this should be solved outside pycparser.

pycparser itself should not be used in `-OO` mode (I'll add a note to the README to this effect). It just makes no sense (just like its dependency PLY). If a dependent project needs to be installed in `-OO` mode somehow, they should not instantiate pycparser during installation. pycparser won't fail now when docstrings aren't present - it's PLY that fails when trying to build the tables.

l29ah:
https://dev.gajim.org/gajim/gajim/issues/9340#note_189294
How do i handle this error?

LefterisJP:
Hey @eliben a tangential question. Where is CFFI's issue tracker where this is tracked? I surprisingly enough can't find it easily

Issue #213, opened 2017-10-06
Consider generating enums on multiple lines.


altendky:
Presently I get:

```
enum ArrayName_e {ArrayName_ParameterName = 0, ArrayName_NewArrayParameterElement = 1, ArrayName_NewArrayParameterElement = 2, ArrayName_NewArrayParameterElement = 3, ArrayName_NewArrayParameterElement = 4, ArrayName_Count = 5};
```

I would personally prefer if enumerations were generated with one element per line such as:

```
enum ArrayName_e {
  ArrayName_ParameterName = 0,
  ArrayName_NewArrayParameterElement = 1,
  ArrayName_NewArrayParameterElement = 2,
  ArrayName_NewArrayParameterElement = 3,
  ArrayName_NewArrayParameterElement = 4,
  ArrayName_Count = 5
};
```

If there is interest in this change I could presumably provide a PR.

eliben:
No objections

altendky:
Thanks once again for all your libs.

PR #216, opened 2017-10-12
Format enums with one value per line

altendky:
Issue #213

### DIFF
--- a/pycparser/c_generator.py
+++ b/pycparser/c_generator.py
@@ -292,7 +292,8 @@ class CGenerator(object):
         if name in ('struct', 'union'):
             members = n.decls
             body_function = self._generate_struct_union_body
-        elif name in ('enum',):
+        else:
+            assert name == 'enum'
             if n.values is None:
                 members = ()
             else:
--- a/pycparser/c_generator.py
+++ b/pycparser/c_generator.py
@@ -135,18 +135,7 @@ class CGenerator(object):
         return ', '.join(visited_subexprs)
 
     def visit_Enum(self, n):
-        s = 'enum'
-        if n.name: s += ' ' + n.name
-        if n.values:
-            s += ' {'
-            for i, enumerator in enumerate(n.values.enumerators):
-                s += enumerator.name
-                if enumerator.value:
-                    s += ' = ' + self.visit(enumerator.value)
-                if i != len(n.values.enumerators) - 1:
-                    s += ', '
-            s += '}'
-        return s
+        return self._generate_struct_union_enum(n, name='enum')
 
     def visit_FuncDef(self, n):
         decl = self.visit(n.decl)
@@ -268,13 +257,13 @@ class CGenerator(object):
         return '...'
 
     def visit_Struct(self, n):
-        return self._generate_struct_union(n, 'struct')
+        return self._generate_struct_union_enum(n, 'struct')
 
     def visit_Typename(self, n):
         return self._generate_type(n.type)
 
     def visit_Union(self, n):
-        return self._generate_struct_union(n, 'union')
+        return self._generate_struct_union_enum(n, 'union')
 
     def visit_NamedInitializer(self, n):
         s = ''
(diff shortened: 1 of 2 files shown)


altendky:
I'm not particularly confident that the punctuation parameter is the proper approach.  Perhaps I should call `self.visit()` directly on the assignment in `_generate_enum_body()`?

altendky:
Incomplete tests, fixing them and fixing a bug with 'empty' enums such as used in typedefs.

tests/test_c_generator.py
```diff
@@ -280,6 +281,25 @@ def test_compound_literal(self):
         self._assert_ctoc_correct('int i = ++(int){ 1 };')
         self._assert_ctoc_correct('struct foo_s foo = (struct foo_s){ 1, 2 };')
 
+    def test_enum(self):
+        s = textwrap.dedent(r'''
```
eliben: Is this dedent necessary? none of the other tests use it since the code is being parsed anyway

pycparser/c_generator.py
```diff
@@ -289,23 +278,55 @@ def visit_NamedInitializer(self, n):
...
-        """ Generates code for structs and unions. name should be either
-            'struct' or union.
+    def _generate_struct_union_enum(self, n, name):
+        """ Generates code for structs, unions, and enums. name should be
+            'struct', 'union', or 'enum'.
         """
+        if name in ('struct', 'union'):
+            members = n.decls
+            body_function = self._generate_struct_union_body
+        elif name in ('enum',):
```
eliben: this should probably be:

```
else:
    assert name == 'enum'
    ... rest of code
```

pycparser/c_generator.py
```diff
@@ -289,23 +278,55 @@ def visit_NamedInitializer(self, n):
...
+            if n.values is None:
+                members = ()
```
eliben: `members = () if n.values is None else n.values.enumerators`

pycparser/c_generator.py
```diff
@@ -289,23 +278,55 @@ def visit_NamedInitializer(self, n):
...
+    def _generate_enum_body(self, members):
+        s = []
+
+        punctuation = ','
+
+        for i, value in enumerate(members):
+            if i == len(members) - 1:
+                punctuation = ''
+
+            s.append(self._generate_stmt(
```
eliben: I wonder why you need to build an AST node again here, why won't just emitting the text do?

tests/test_c_generator.py
altendky: `_assert_ctoc_correct()` is not checking the formatting and my change is a formatting change so it would seem that my test should do at least something different.  I just verified this by checking that all three of the following tests pass.

```
    def test_enum_a(self):
        s = textwrap.dedent(r'''
            enum e
            {
              a = 1,
              b = 2,
              c = 3
            };
        '''[1:])

        self._assert_ctoc_correct(s)

    def test_enum_b(self):
        s = r'''
            enum e
            {
              a = 1,
              b = 2,
              c = 3
            };
        '''

        self._assert_ctoc_correct(s)

    def test_enum_c(self):
        s = r'''enum e {a = 1, b = 2, c = 3};'''

        self._assert_ctoc_correct(s)
```

pycparser/c_generator.py
altendky: It is my understanding that `assert` is only proper in test code.  But, the gist that an exception should be raised for an invalid input does sound good.  How about leaving the `elif` and adding an `else` that raises a to-be-defined exception?

```
UnknownTypeError('`{name}` not recognized'.format(name=name))
```

Given that this is an internal method it seems that we may not need a nice user error.  I'll note that as is we'd get an `UnboundLocalError` for the following access to `members`.  I added a test for that and will adjust it based on the implementation decision.

pycparser/c_generator.py
altendky: Pending above possibilities I'll change this.

pycparser/c_generator.py
altendky: I was trying to make it look like how `struct` and `union` were done, but failing.  I removed the punctuation parameter from `_generate_stmt()` and added `visit_Enumerator()`.  Perhaps this will fit better with the existing code.

pycparser/c_generator.py
eliben: I actually think `assert` is fine in this case, since it's checking an invariant in the code - it's a bug if this place is reached, so `assert` is appropriate.

tests/test_c_generator.py
eliben: Ah, I see. If that's the case, no problem. The other tests don't tend to do one-to-one comparison of actual formatting, but I guess it's OK in this case. I'm a bit worried that this test will be fragile because it compares large strings, but we can fix it if it becomes and issue.

pycparser/c_generator.py
```diff
@@ -289,22 +285,44 @@ def visit_NamedInitializer(self, n):
...
             self.indent_level -= 2
             s += self._make_indent() + '}'
         return s
 
+    def _generate_struct_union_body(self, members):
+        return ''.join(self._generate_stmt(decl) for decl in members)
+
+    def _generate_enum_body(self, members):
+        s = []
+
```
eliben: Python usually prefers denser style - none of the empty lines in this function are really necessary

pycparser/c_generator.py
```diff
@@ -289,22 +285,44 @@ def visit_NamedInitializer(self, n):
...
+        for value in members:
```
eliben: Actually, this loop can be turned into a simple list comprehension, no?

pycparser/c_generator.py
```diff
@@ -289,22 +285,44 @@ def visit_NamedInitializer(self, n):
...
+            line = self.visit(value)
+
+            s.append(line)
+
+        return ''.join(s)[:-2] + '\n'
```
eliben: A comment should say why the `[:-2]` is needed here

PR #218, opened 2017-10-27
Fixup C Preprocessor

laerreal:
There are several errors in macro expansion algorithm. 
This patch series first adds example files for each error.
Then it suggests fixes.


laerreal:
GitHub mixed this patches up because of inconsequent timestamps after rebasing...

eliben:
This modifies PLY - a dependency of pycparser; I suggest you send a PR to PLY instead

Issue #219, opened 2017-11-15
Improve NodeVisitor performance


ldore:
I'm using `NodeVisitor`s to visit ASTs on large source files (1MB+).

When profiling my scripts, a significant part of the execution time is spent 
- in the `Node.children()` method, generating the name/value tuples
- in the `NodeVisitor.visit()` method, computing the name and looking up the actual visitor method.

Performance could be improved by:
- Providing an iterator on the children, without creating intermediate lists/tuples objects. 
  Names could be omitted as they are not used by `NodeVisitor`.
- Using an internal cache of `visit_*` functions to speed up their lookup.

ldore:
See pull request #220

myint:
Closed by #220.

PR #220, opened 2017-11-15
Improve NodeVisitor performance, add iterator on Node children (Issue #219).

ldore:
This is a resolution for Issue #219 
Performance improvement can be measured with the following script:
```
with self._open_c_file('cppd_with_stdio_h.c') as f:
    code = f.read()
setup = '''
from pycparser.c_ast import NodeVisitor
from pycparser.c_parser import CParser
code = %r
p = CParser().parse(code)
visitor = NodeVisitor()
''' % (code)
print(timeit.timeit('visitor.visit(p)',
                     setup = setup,
                     number=100))
```

On my workstation, this is twice as fast as the original implementation, while passing all unit tests.

### DIFF
--- a/pycparser/_ast_gen.py
+++ b/pycparser/_ast_gen.py
@@ -84,6 +84,7 @@ class NodeCfg(object):
     def generate_source(self):
         src = self._gen_init()
         src += '\n' + self._gen_children()
+        src += '\n' + self._gen_iter()
         src += '\n' + self._gen_attr_names()
         return src
 
@@ -128,6 +129,35 @@ class NodeCfg(object):
             src += '        return tuple(nodelist)\n'
         else:
             src += '        return ()\n'
+
+        return src
+
+    def _gen_iter(self):
+        src = '    def __iter__(self):\n'
+
+        if self.all_entries:
+            for child in self.child:
+                src += (
+                    '        if self.%(child)s is not None:\n' +
+                    '            yield self.%(child)s\n') % (dict(child=child))
+
+            for seq_child in self.seq_child:
+                src += (
+                    '        for child in (self.%(child)s or []):\n'
+                    '            yield child\n') % (dict(child=seq_child))
+                    
+            if not (self.child or self.seq_child):
+                # Empty generator
+                src += (
+                    '        return\n' +
+                    '        yield\n'
+                    )
+        else:
+            # Empty generator
+            src += (
+                '        return\n' +
+                '        yield\n'
+                )
 
         return src
 
@@ -253,21 +283,35 @@ class NodeVisitor(object):
         *   Modeled after Python's own AST visiting facilities
             (the ast module of Python 3.0)
     """
+    
+    _method_cache = None
+    
     def visit(self, node):
         """ Visit a node.
         """
-        method = 'visit_' + node.__class__.__name__
-        visitor = getattr(self, method, self.generic_visit)
+        
+        method_cache = self._method_cache
+        if method_cache is None:
+            method_cache = {}
+            self._method_cache = method_cache
+            
+        visitor = method_cache.get(node.__class__.__name__, None)
+        if visitor is None:
+            method = 'visit_' + node.__class__.__name__
+            visitor = getattr(self, method, self.generic_visit)
+            
+            method_cache[node.__class__.__name__] = visitor
+            
         return visitor(node)
 
     def generic_visit(self, node):
         """ Called if no explicit visitor function exists for a
             node. Implements preorder visiting of the node.
         """
-        for c_name, c in node.children():
+        for c in node:
             self.visit(c)
-
-
+            
+            
 '''
 
 
--- a/pycparser/_ast_gen.py
+++ b/pycparser/_ast_gen.py
@@ -290,16 +290,14 @@ class NodeVisitor(object):
         """ Visit a node.
         """
 
-        method_cache = self._method_cache
-        if method_cache is None:
-            method_cache = {}
-            self._method_cache = method_cache
+        if self._method_cache is None:
+            self._method_cache = {}
 
-        visitor = method_cache.get(node.__class__.__name__, None)
+        visitor = self._method_cache.get(node.__class__.__name__, None)
         if visitor is None:
             method = 'visit_' + node.__class__.__name__
             visitor = getattr(self, method, self.generic_visit)
-            method_cache[node.__class__.__name__] = visitor
+            self._method_cache[node.__class__.__name__] = visitor
 
         return visitor(node)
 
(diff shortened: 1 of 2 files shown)


pycparser/_ast_gen.py
```diff
@@ -253,21 +283,35 @@ def visit_Constant(self, node):
...
+    
+    _method_cache = None
+    
     def visit(self, node):
         """ Visit a node.
         """
-        method = 'visit_' + node.__class__.__name__
-        visitor = getattr(self, method, self.generic_visit)
+        
+        method_cache = self._method_cache
```
eliben: What's the reason for having `method_cache` here locally? Are you trying to cache the access to the `self._method_cache` attribute? I wouldn't expect this to change performance much - can you check?

pycparser/_ast_gen.py
```diff
@@ -253,21 +283,35 @@ def visit_Constant(self, node):
...
+        if method_cache is None:
+            method_cache = {}
+            self._method_cache = method_cache
+            
+        visitor = method_cache.get(node.__class__.__name__, None)
+        if visitor is None:
+            method = 'visit_' + node.__class__.__name__
+            visitor = getattr(self, method, self.generic_visit)
+
```
eliben: This empty line is unnecessary

pycparser/_ast_gen.py
```diff
@@ -253,21 +283,35 @@ def visit_Constant(self, node):
...
         """
-        method = 'visit_' + node.__class__.__name__
-        visitor = getattr(self, method, self.generic_visit)
+        
+        method_cache = self._method_cache
+        if method_cache is None:
+            method_cache = {}
+            self._method_cache = method_cache
+            
+        visitor = method_cache.get(node.__class__.__name__, None)
```
eliben: Can you use dict's `setdefault` here?

pycparser/_ast_gen.py
```diff
@@ -253,21 +283,35 @@ def visit_Constant(self, node):
...
     def generic_visit(self, node):
         """ Called if no explicit visitor function exists for a
             node. Implements preorder visiting of the node.
         """
-        for c_name, c in node.children():
+        for c in node:
             self.visit(c)
-
-
+
```
eliben: Remove trailing whitespace here and elsewhere

pycparser/_ast_gen.py
ldore: No, that's not for caching access, but for lazy initialization of the `_method_cache` attribute.
The method cache should be specific to each `NodeVisitor`-derived type.

To do that, we could:
- Use a class attribute. 
  If we do that, then everyone implementing a `NodeVisitor`-derived class would have to define `_method_cache` class attribute, initialized to `{}`.
  As that's not very user-friendly, and would break existing implementations, I didn't do it.
- Use an instance attribute.
  The pythonic way would define the instance attribute with `self._method_cache = {}` in `NodeVisitor.__init__()`. If we do that, then everyone implementing a `NodeVisitor`-derived class has to call the parent constructor.
  For backward compatibility, I didn't do it.
- Use `__metaclass__`.
  Metaclasses are a nice way to customize the creation of a class, and would serve the purpose well.
  Unfortunately, they go a bit beyond simple python, and make the code a bit less maintainable.
  So I didn't use them either (I tested it, though, and can submit them).

What was implemented:
- I defined a default, `None`-initialized `_method_cache` at class level, which is only read.
  When executing `visit` for the first time, and reading `self._method_cache`, we actually read `self.__class__._method_cache`, which is `None`.
  When we set `self._method_name` to a default, empty `dict`, this effectively creates `self._method_name` at instance level.
  As `self._method_name` now exists, further reads will return the instance value, not the class value.
  I chose to not set `self.__class__._method_cache`, as this could have caused problems for users maintaining a class hierarchy of `NodeVisitor`.

For the current implementation:
Pros:
 - This ensures that the variable exists, even if the writer of a `NodeVisitor`-derived class forgot to define a `method_cache` class variable, or forgot to call the parent constructor.
 - This is backward-compatible with existing implementations (indeed, all tests still pass)
Cons:
 - This creates a `_method_cache` variable at instance level, where it would be better set at class level.
   This can be fixed by using `self.__class__` instead of `self`.
 - This is not very pythonic...

I did some additional performance checks (focused on the cache mechanism, and not using the generator):
```
10000 iterations: 15.567s on Original
10000 iterations: 11.181s on Method cache only, using class name as key (pull request contents)
10000 iterations: 10.957s on Method cache only, using class type as key
10000 iterations: 10.903s on Method cache only, define cache at __init__
10000 iterations: 10.785s on Method cache only, define cache in metaclass
```

pycparser/_ast_gen.py
ldore: That would be functionally equivalent, but performance would be similar to the original implementation, as we would still compute/retrieve the visitor for each call.

Using `collections.defaultdict` is closer to the intended improvement: It creates a value only when the key is missing. However, it requires deriving a class from `defaultdict` and implementing `__missing__`.
As we need to access the current class' attributes, we need to either store a reference to it in the class, or use a closure.

I implemented both, and was surprised to see no improvement:
```
10000 iterations: 15.272s on Original
10000 iterations: 11.264s on Method cache only, using class name as key
10000 iterations: 17.910s on Method cache only, using setdefault
10000 iterations: 11.855s on Method cache only, using defaultdict
```

So `setdefault` is not better, and neither is `defaultdict`.

pycparser/_ast_gen.py
ldore: Should be fixed now.

pycparser/_ast_gen.py
ldore: Should be fixed now.

pycparser/_ast_gen.py
eliben: Thanks for the detailed reply. I think I understood the lazy initialization part. My question was about this code:

+        method_cache = self._method_cache
+        if method_cache is None:
+            method_cache = {}
+            self._method_cache = method_cache

You have a local `method_cache` here, along with `self._method_cache`. My question was why isn't this code just:

````
if self._method_cache is None:
    self._method_cache = {}
````

And using `self._method_cache` directly instead of `method_cache` afterwards.

pycparser/_ast_gen.py
eliben: OK, thanks

pycparser/_ast_gen.py
ldore: That's an oversight, thanks for spotting it.
Removing the local variable, and using `self._method_cache` directly is both more concise and a bit faster.

Issue #221, opened 2017-11-22
Support #pragma in struct declarations


ldore:
Pragmas do not seem to be supported within typedefs.
Updating `tests.test_c_generator.TestCtoC.test_pragma` to: 
```
    def test_pragma(self):
        self._assert_ctoc_correct(r'''
            #pragma foo
            void f() {
                #pragma bar
                i = (a, b, c);
            }
            typedef struct s {
            #pragma baz
            } s;
        ''')
```
raises a `ParseError`.

This occured in a source similar to:
```
typedef struct MyStruct {
    int a;
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
    int b;
#pragma GCC diagnostic pop
}
```

myint:
Perhaps this should be closed (#222)?

PR #222, opened 2017-11-22
Add support for #pragma in struct_declaration (Issue #221).

ldore:
Support `#pragma` within a `struct`

Note: On first run (when `lextab.py`/`yacctab.py` do not exist yet), I get the following message:
```
Generating LALR tables
WARNING: 1 shift/reduce conflict
```
As this also happens on master, I assume this is OK.


eliben:
Merging, thanks!

Issue #226, opened 2017-12-14
Support __repr__ on nodes


ldore:
When creating parts of an AST from scratch, figuring out the actual types to use (as well as their correct constructor parameters), is currently a manual (and error-prone), process:
1. Create a source close to the intended code
2. Dump the AST, using `FileAST.show()`
3. Guess what types are needed, and refine until the correct structure is achieved.

Implementing [`__repr__`](https://docs.python.org/3/library/functions.html#repr) would definitely help, as step 3 would be a copy/paste of the `__repr__` output.

For example, using `tests_list_children`'s structure, `repr(comp)` could return:
```
Compound(block_items=[BinaryOp(op='+',
                               left=Constant(type='float',
                                             value='5.6'
                                             ),
                               right=Constant(type='char',
                                              value='t'
                                              )
                               ),
                      BinaryOp(op='-',
                               left=BinaryOp(op='+',
                                             left=Constant(type='float',
                                                           value='5.6'
                                                           ),
                                             right=Constant(type='char',
                                                            value='t'
                                                            )
                                             ),
                               right=Constant(type='char',
                                              value='t'
                                              )
                               ),
                      Constant(type='float',
                               value='5.6'
                               ),
                      Constant(type='char',
                               value='t'
                               )
                     ]
         )
```

eliben:
I'm not sure I understand the problem you're describing. Have you tried the `attrnames` and `nodenames` parameters to `show` for more visibility?

ldore:
Using `attrnames=True` and `nodenames=True` in `show` indeed gives more _visibility_.
As I often build (parts of an) AST manually (rewrites on massive C source base), this still leaves a lot of manual conversion to python code.
Part of the python doc for [`__repr__`](https://docs.python.org/3/library/functions.html#repr) states:
> ...  this function makes an attempt to return a string that would yield an object with the same value when passed to eval() ...

This is what I would like to achieve: A function that produces python code that can be copy/pasted to duplicate (parts of) the AST.

I rewrote and simplifed the `__repr__` implementation, so that it only relies on `__slots__`: changes to pycparser are now a lot smaller.

eliben:
OK, I see. I've left a review on the PR.

PR #227, opened 2017-12-14
Implement __repr__ on Nodes (Issue #226)

ldore:
This is a first attempt at implementing python's `__repr__` method on `*Node` types (Issue #226).

Notes:
- This is limited to constructor parameters (any attribute not passed in constructor parameters is not displayed).
- The `coord` constructor parameter is not displayed. This is intentional, as my main use-case of the `__repr__` method is for rapid prototyping of an AST, and `coord` would just add noise to the output.
- I don't really like the way `_get_repr` is implemented (a bit too cryptic), or the way generated code works (creates a lot of intermediate objects, and does too many string replacements).
- Style of the results of a `repr` call tries to be close to PEP8.

Any improvement idea is welcome.

### DIFF
--- a/pycparser/_ast_gen.py
+++ b/pycparser/_ast_gen.py
@@ -86,7 +86,7 @@ class NodeCfg(object):
         src = self._gen_init()
         src += '\n' + self._gen_children()
         src += '\n' + self._gen_iter()
-        src += '\n' + self._gen_repr()
+
         src += '\n' + self._gen_attr_names()
         return src
 
@@ -161,21 +161,6 @@ class NodeCfg(object):
 
         return src
 
-    def _gen_repr(self):
-        src = '    def __repr__(self):\n'
-        src += '        result = "%s("\n' % self.name
-        for entry in self.all_entries:
-            src += '        result += "%s=%%s%s" %% (_repr(self.%s).replace("\\n", "\\n%s"))\n' % (entry,
-                                                                                                   (',\\n' if entry != self.all_entries[-1] else '\\n'),
-                                                                                                   entry,
-                                                                                                   ' ' * (len(entry) + 1))
-
-        src += '        result += ")"\n'
-        src += '        result = result.replace("\\n", "\\n%s")\n' % (' ' *
-                                                                      (len(self.name) + 1))
-        src += '        return result\n'
-        return src
-
     def _gen_attr_names(self):
         src = "    attr_names = (" + ''.join("%r, " % nm for nm in self.attr) + ')'
         return src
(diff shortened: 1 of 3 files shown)


pycparser/_ast_gen.py
```diff
@@ -187,11 +189,38 @@ def _gen_attr_names(self):
 _PROLOGUE_CODE = r'''
 import sys
 
+def _repr(obj):
+    """
+    Get the representation of an object, with dedicated pprint-like format for lists and tuples.
+    """
+    if isinstance(obj, list):
```
eliben: Why only list if the comment says tuples too?

pycparser/_ast_gen.py
```diff
@@ -187,11 +189,38 @@ def _gen_attr_names(self):
...
+        return '[' + (',\n '.join((_repr(e).replace('\n', '\n ') for e in obj))) + '\n]'
```
eliben: Why the `\n` replacement adding whitespace?

pycparser/_ast_gen.py
```diff
@@ -187,11 +189,38 @@ def _gen_attr_names(self):
...
     """ Abstract base class for AST nodes.
     """
+    def __repr__(self):
+        """ Generates a python representation of the current node
+        """
+        result = self.__class__.__name__ + '('
+        
+        indent = ''
+        separator = ''
+        for name in self.__slots__[:-2]:
```
eliben: Is there a reason to use `__slots__` directly here? I mean why not just use `chldren` and `attr_names`?

pycparser/_ast_gen.py
ldore: This is for adding indentation for parameters, so that they are aligned with their opening delimiter.
The result is similar to the first of [PEP8's indentation examples](https://www.python.org/dev/peps/pep-0008/#id3), each parameter except the first on a new line.

Without it, a `BinaryOp(op='+', left=Constant(type='int', value='1'), right=Constant(type='int', value='2'))` would look like:
```
BinaryOp(op='+', 
         left=Constant(type='int',
         value='1'),
         right=Constant(type='int',
         value='2')
         )
```
instead of:
```
BinaryOp(op='+', 
         left=Constant(type='int',
                       value='1'),
         right=Constant(type='int',
                        value='2')
         )
```

pycparser/_ast_gen.py
ldore: Here, `children` would not work for sequences:  `Compound.block_items` would be returned as a list of pairs `('block_item[%d]', <node>)`. 
Infering the parameters from the list of tuples, while possible, would be a lot less efficient (and more error-prone).

As for `attr_names`, this reflects the instance attributes, not the constructor parameters.

To sum it up, using `Compound` as an example:
- `Compound.__slots__` contains `('block_items', 'coord', '__weakref__')`
- `Compound.attr_names` contains `()`
- `Compound.children()` returns `[('block_items[0]', <node_0>), ... , ('block_items[n]', <node_n>)]`

pycparser/_ast_gen.py
ldore: My first implementation handled `tuple`s too.
Then I realized that only `list`s were created by `c_parser`, and that I only ever used `list`s too.
So I dropped the `tuple`-specific code (but forgot to remove the comment).

PR #228, opened 2017-12-16
Add missing fake_libc_includes for POSIX.1-2008 compatibility

bclephas:
Various fake libc includes were missing according the POSIX.1-2008 standard. Add those.
See http://pubs.opengroup.org/onlinepubs/9699919799/ and specifically http://pubs.opengroup.org/onlinepubs/9699919799/idx/head.html


eliben:
Thanks!

PR #232, opened 2018-01-22
Extend Xlib objects in fake includes

Alexnagel:
Extend the Xlib objects. Needed for a package using a wider range of Xlib items.


Alexnagel:
Copied wrong file, missing one definition in this commit

PR #233, opened 2018-01-22
Extend Xlib objects in fake includes

Alexnagel:
Extend the Xlib objects. Needed for a package using a wider range of Xlib items.


Alexnagel:
I went forward on the work of #180, that's why the separate files. On second thought those function declarations aren't needed indeed, just checked. Good call.

A solution for collisions I see is creating _fake_x.h in the X11 directory containing the X11 types. Would this be a viable solution?

eliben:
Yes, I think separating the x11 thing into its own header would be good. I'm just concerned that the "fake typedefs" file gets included by all headers, but if this goes separately I don't think it will be problematic.

Feel free to refactor the code from #180 to move *all* X11 stuff into separate headers

Alexnagel:
I have updated the change to separate the X11 items from the general fakes and removed the unnecessary function declarations

eliben:
Thanks!

PR #236, opened 2018-03-01
Fix #235: Pragma displacing real statements

### LINKED ISSUE #235: Incorrect AST Structure when #pragma follows For loop.
Consider the following code snippet: 
```c
for(int i = 0; i < 3; i++)
    #pragma omp critical
        sum += 1;
```
When compiled without Open MP, the #pragma should be ignored completely, so that the statement ``sum += 1`` should be a descendent of the for loop. However, in the current implementation of pycparser, it is parsed a _sibling_ of the for loop, instead of as a descendant. 
```
For: 
  DeclList: 
    Decl: i, [], [], []
      TypeDecl: i, []
        IdentifierType: ['int']
      Constant: int, 0
  BinaryOp: <
    ID: i
    Constant: int, 3
  UnaryOp: p++
    ID: i
  Pragma: omp critical
Assignment: +=
  ID: sum
  Constant: int, 1
```

The same problem applies to other loops, Labels, and If statements, as in the following:
```c
for(int i = 0; i < 3; i++)
    myLabel:
    #pragma omp critical
        sum += 1;
```
```c
while(sum < 100)
    #pragma omp critical
        sum += 1;
```
```c
if (sum < 100)
    #pragma omp critical
    sum += 1;
```
The following will not even parse, but it should:
```c        
do 
    #pragma omp critical
    sum += 1;
while(sum < 100)
```

dbluhm:
I apologize if this is terribly naive but I believe this should take care of #235 by wrapping pragma's and the directly following statement in a Compound. As discussed in #235 this is kind of weird but appears to be the best way to ensure (mostly) correct AST structure in these cases.

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -613,8 +613,9 @@ class CParser(PLYParser):
                         | iteration_statement
                         | jump_statement
                         | pppragma_directive statement
+                        | pppragma_directive
         """
-        if isinstance(p[1], c_ast.Pragma):
+        if isinstance(p[1], c_ast.Pragma) and len(p) == 3:
             p[0] = c_ast.Compound(
                 block_items=[p[1], p[2]],
                 coord=self._token_coord(p, 1))
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -612,8 +612,13 @@ class CParser(PLYParser):
                         | selection_statement
                         | iteration_statement
                         | jump_statement
-                        | pppragma_directive statement
                         | pppragma_directive
+        """
+        p[0] = p[1]
+
+    def p_pragmacomp_or_statement(self, p):
+        """ pragmacomp_or_statement	: pppragma_directive statement
+                                        | statement
         """
         if isinstance(p[1], c_ast.Pragma) and len(p) == 3:
             p[0] = c_ast.Compound(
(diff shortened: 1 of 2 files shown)


dbluhm:
I failed to run tests before submitting :stuck_out_tongue_closed_eyes:. I'm looking into the errors reported now.

dbluhm:
One issue was that the there was no default case for pragma's without statements. Now, however, tests fail due to the structure being slightly different. @eliben, what do you think of these changes?

dbluhm:
Narrowed the affected pragma's with b1ff55d

pycparser/c_parser.py
```diff
@@ -616,6 +616,17 @@ def p_statement(self, p):
         """
         p[0] = p[1]
 
+    def p_pragmacomp_or_statement(self, p):
+        """ pragmacomp_or_statement	: pppragma_directive statement
+                                        | statement
```
eliben: Align the | with the : on the line above

pycparser/c_parser.py
```diff
@@ -616,6 +616,17 @@ def p_statement(self, p):
         """
         p[0] = p[1]
 
+    def p_pragmacomp_or_statement(self, p):
+        """ pragmacomp_or_statement	: pppragma_directive statement
```
eliben: Can the pragma come after the statement?

pycparser/c_parser.py
```diff
@@ -616,6 +616,17 @@ def p_statement(self, p):
         """
         p[0] = p[1]
 
+    def p_pragmacomp_or_statement(self, p):
```
eliben: This needs a comment explaining what's going on, and also tests

pycparser/c_parser.py
dbluhm: I can't think of a situation where a statement followed by a pragma would be a case for grouping with a compound. The pragma is the decorator for the following statement but the statement preceding a pragma wouldn't mean anything to the pragma.

Issue #240, opened 2018-03-08
Valueless enum members are generated incorrectly


natezb:
This seems to have been introduced by #216, which inadvertently removed some special-case checks. Enum members that don't have an assigned value now generate an invalid, empty assignment. For example:
```
typedef enum MOT_TravelModes
{
  MOT_TravelModeUndefined = ,
  MOT_Linear = 0x01,
  MOT_Rotational = 0x02
} MOT_TravelModes;
```
In the original code, `MOT_TravelModeUndefined` was assigned no value, and should be generated with no `=` sign.

SethPoulsen:
I just realized the same problem.

eliben:
@altendky PTAL

SethPoulsen:
No need. I got it sorted out.

saullocarvalho:
Is this still an issue? It seems like a commit has already fixed it.

SethPoulsen:
Ya, I fixed it. It should be closed.

natezb:
Just tested under 2.19, and works for me.

PR #243, opened 2018-03-15
Fixed enum formatting in generating C code (issue #240).

SethPoulsen:


SethPoulsen:
Done. I also took the liberty of condensing the test_casts function, since there were two functions defined with the same name.

pycparser/c_generator.py
```diff
@@ -138,11 +138,17 @@ def visit_Enum(self, n):
         return self._generate_struct_union_enum(n, name='enum')
 
     def visit_Enumerator(self, n):
-        return '{indent}{name} = {value},\n'.format(
-            indent=self._make_indent(),
-            name=n.name,
-            value=self.visit(n.value),
-        )
+        if not n.value:
```
Syeberman: Is `n.value` an `int`, or a `str`?  If it's an `int`, `n.value` will be false if it's zero.

eliben: Isn't `n.value` a node?

SethPoulsen: That's right, its a node (pycparser.c_ast.Constant to be exact). So it will only be false if there is no node there, which is what we want.

PR #245, opened 2018-03-26
Fixing redefinition in Fake Headers.

SethPoulsen:
Currently, both `_fake_typedefs.h` and `_fake_defines.h` define `__builtin_va_list` and `va_list`.
A case could be made for having them in either file. I don't really care which file they are defined in, I just don't want it to be in both of them, because its a redefinition error.

### DIFF
--- a/utils/fake_libc_include/_fake_typedefs.h
+++ b/utils/fake_libc_include/_fake_typedefs.h
@@ -4,6 +4,7 @@
 typedef int size_t;
 typedef int __builtin_va_list;
 typedef int __gnuc_va_list;
+typedef int va_list;
 typedef int __int8_t;
 typedef int __uint8_t;
 typedef int __int16_t;
--- a/utils/fake_libc_include/_fake_typedefs.h
+++ b/utils/fake_libc_include/_fake_typedefs.h
@@ -2,6 +2,7 @@
 #define _FAKE_TYPEDEFS_H
 
 typedef int size_t;
+typedef int __builtin_va_list;
 typedef int __gnuc_va_list;
 typedef int __int8_t;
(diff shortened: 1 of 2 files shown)


utils/fake_libc_include/_fake_typedefs.h
```diff
@@ -2,7 +2,6 @@
 #define _FAKE_TYPEDEFS_H
 
 typedef int size_t;
-typedef int __builtin_va_list;
```
eliben: Is this one duplicated too?

SethPoulsen: Ah, no its not. Its being used to define a different macro. Thanks for catching my whoops. However, since `__builtin_va_list` ends up being parsed as a function call, it doesn't matter if its defined in the fake headers.

SethPoulsen: Sorry, my bad again. I was mixing up `__builtin_va_list` with `__builtin_va_arg`... so many macros...

utils/fake_libc_include/_fake_typedefs.h
```diff
@@ -150,8 +150,6 @@ typedef int uintmax_t;
 /* C99 stdbool.h bool type. _Bool is built-in in C99 */
 typedef _Bool bool;
 
-typedef int va_list;
```
eliben: Sorry, I think it'd be actually better to keep this in this file and remove from the other. It's a typedef, after all... and it's needed to parse types, whereas fake_typedefs is needed for the preprocessor

SethPoulsen: No problem. Like I said, I have no preference.

PR #254, opened 2018-04-26
Add support for empty struct (#66)

ldore:
Support empty `struct` in parser and c generator.

Though not allowed by C99, empty `struct`s can be found in android headers.
(Look for `__bionic_zero_size_is_okay_t` in android code).

Note: 
Issue #66 was closed because the original poster found a workaround. 
As @eliben welcomed pull requests in comments, #66 could be reopened/reclosed as fixed.

### DIFF
--- a/pycparser/c_generator.py
+++ b/pycparser/c_generator.py
@@ -300,10 +300,12 @@ class CGenerator(object):
             body_function = self._generate_struct_union_body
         else:
             assert name == 'enum'
-            members = () if n.values is None else n.values.enumerators
+            members = None if n.values is None else n.values.enumerators
             body_function = self._generate_enum_body
         s = name + ' ' + (n.name or '')
         if members is not None:
+            # None means no members
+            # Empty sequence means an empty list of members
             s += '\n'
             s += self._make_indent()
             self.indent_level += 2
--- a/pycparser/c_generator.py
+++ b/pycparser/c_generator.py
@@ -303,7 +303,7 @@ class CGenerator(object):
             members = () if n.values is None else n.values.enumerators
             body_function = self._generate_enum_body
         s = name + ' ' + (n.name or '')
-        if members:
+        if members is not None:
             s += '\n'
             s += self._make_indent()
             self.indent_level += 2
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -908,6 +908,7 @@ class CParser(PLYParser):
                                         | struct_or_union TYPEID
         """
         klass = self._select_struct_union_class(p[1])
+        # None means no list of members
         p[0] = klass(
             name=p[2],
             decls=None,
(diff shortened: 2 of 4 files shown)


pycparser/c_parser.py
```diff
@@ -923,14 +923,23 @@ def p_struct_or_union_specifier_2(self, p):
...
+        if len(p) == 6:
+            p[0] = klass(
+                name=p[2],
+                decls=p[4],
+                coord=self._token_coord(p, 2))
+        else:
+            # Empty structs are not pure-C99, but accepted.
+            p[0] = klass(
+                name=p[2],
+                decls=[],
```
eliben: Worth mentioning in a comment that a distinction is now made between None and `[]` for the `decls`?

eliben:
Thanks! Please fix the failures to proceed

ldore:
Fixed, and slightly simplified the grammar.

pycparser/c_parser.py
```diff
@@ -914,31 +915,39 @@ def p_struct_or_union_specifier_1(self, p):
...
-        """ struct_or_union_specifier : struct_or_union brace_open struct_declaration_list brace_close
+        """ struct_or_union_specifier : struct_or_union brace_open struct_declaration_list_opt brace_close
         """
         klass = self._select_struct_union_class(p[1])
-        p[0] = klass(
-            name=None,
-            decls=p[3],
-            coord=self._token_coord(p, 2))
+        if len(p) == 4:
+            # None means no members
```
eliben: Seems like the code has changed, there's no None assignment now - is this comment still relevant?

pycparser/c_generator.py
```diff
@@ -300,10 +300,12 @@ def _generate_struct_union_enum(self, n, name):
             body_function = self._generate_struct_union_body
         else:
             assert name == 'enum'
-            members = () if n.values is None else n.values.enumerators
+            members = None if n.values is None else n.values.enumerators
             body_function = self._generate_enum_body
         s = name + ' ' + (n.name or '')
         if members is not None:
+            # None means no members
```
eliben: What does this comment refer to now, in the current revision of the code?

pycparser/c_parser.py
ldore: Still relevant, but not in this function.
Comment moved to `p_struct_or_union_specifier_1` (which handles the `None` case)

pycparser/c_generator.py
ldore: Same thing as before.
- `None` still means that no curly-braces-enclosed members are present (handled by `p_struct_or_union_specifier_1` in the parser.
- Non-empty sequence means curly braces with members.
- Empty sequence means curly braces with no members.

Line 303 (translating `None` to an empty `tuple`) caused enums to always be suffixed with (possibly empty) curly braces.

eliben:
Thanks!

Rahim16:
Still cannot parse empty structs in version 2.19, python 2.7. Running this code:
`
a = """
struct ngram_funcs_s 
{
}
;
"""
from pycparser import c_parser
parser = c_parser.CParser()
parser.parse(a)
`
results in:
`
pycparser.plyparser.ParseError: :4:1: before: }
`

PR #255, opened 2018-04-26
Fix non-generated constant expressions in designated initializers (#246)

### LINKED ISSUE #246: Constant expressions in designated initializers are not generated back to C
While pycparser correctly parses a constant-expression in a designated initializer (the AST is correct), it fails to write it back when generating C code.

Consider the following code:
```C
void myFunction(void)
{
  int array[3] = {[0] = 0, [1] = 1, [1+1] = 2};
}
```
Parsing it, then using `CGenerator` to generate the source produces:
```C
void myFunction(void)
{
  int array[3] = {[0] = 0, [1] = 1, = 2};
}
```

The C99 grammar describes the designator part of designated initializers as:
```ebnf
designator: [ constant-expression ]
            . identifier
```
(See §6.7.8 in http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf)

The ```CGenerator.visit_NamedInitializer``` currently only considers the `ID` and `Constant` types.
The `Constant` branch should either be extended to other types or be an `else:` branch.

ldore:
Adds a visitor call on non-`ID` names in `CGenerator.visit_NamedInitializer` (Fix #246).

### DIFF
--- a/pycparser/c_generator.py
+++ b/pycparser/c_generator.py
@@ -283,8 +283,8 @@ class CGenerator(object):
         for name in n.name:
             if isinstance(name, c_ast.ID):
                 s += '.' + name.name
-            elif isinstance(name, c_ast.Constant):
-                s += '[' + name.value + ']'
+            else:
+                s += '[' + self.visit(name) + ']'
         s += ' = ' + self._visit_expr(n.expr)
         return s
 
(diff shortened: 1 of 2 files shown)


eliben:
Thanks!

PR #260, opened 2018-05-20
Replace a call to Popen by check_output in order to check that cpp returns 0.

marmeladema:
At the moment, if cpp fails, no error/exception is generated and pycparser continues.
Calling check_output instead of Popen solves this.


eliben:
Thanks!

PR #266, opened 2018-06-16
Remove unnecessary __future__ import

jdufresne:
Generators have been available since 2.3. The feature is automatically included in all supported Pythons.

For additional details, see:

https://docs.python.org/3/library/__future__.html


eliben:
Thanks for all the contributions

PR #269, opened 2018-06-16
Avoid opening files with deprecated 'U' mode

jdufresne:
Opening files with 'U' mode is deprecated. When running tests with Python warnings enabled, the warnings of the following form are emitted:

```
  DeprecationWarning: 'U' mode is deprecated
    return open(name, 'rU')
```

To open files with universal newlines on both Ptyhon 2 & 3, use the io module. It defaults to opening with universal newlines and doesn't emit a warning.

https://docs.python.org/3/library/io.html

> When reading input from the stream, if newline is None, universal newlines mode is enabled.

### DIFF
--- a/pycparser/__init__.py
+++ b/pycparser/__init__.py
@@ -10,6 +10,7 @@
 __all__ = ['c_lexer', 'c_parser', 'c_ast']
 __version__ = '2.18'
 
+import io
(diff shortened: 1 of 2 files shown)


eliben:
Thanks for the PRs. Just back from vacation - will try to get to these soon

PR #275, opened 2018-07-21
add support for structs/unions with 0 fields.

Sauci:


eliben:
This should already work!
See tests I added in https://github.com/eliben/pycparser/commit/92f65b6bd8539dfe7d3b1e8407d74092e03cb11b

Maybe you had stale lex/yacc tables?

Sauci:
you are right, it should work, and it is (I didn't see this test before writing my own). but when I use the package installed using pip command in an other project, the parser fails on such a code pattern. do you have an idea of what could lead to this issue?

eliben:
Maybe the change didn't make it into a released version yet?

Sauci:
ok, that would explain it... then I guess the pull request should be ignored, as well as the issue?
thanks for your feedback

eliben:
Yes, please close both

Issue #253, opened 2018-04-19
Incorrect Parsing of Floating Point Literals


SethPoulsen:
Currently, the floating point literal
```
double f = 4.5;
```
is parsed to 
```
Decl: f, [], [], []
  TypeDecl: f, []
    IdentifierType: ['double']
  Constant: float, 4.5
```
when really, the `type` attribute of the `Constant` node should be "double". Likewise, 
```
double f = 4.5L;
```
is parsed to 
```
Decl: f, [], [], []
  TypeDecl: f, []
    IdentifierType: ['double']
  Constant: float, 4.5L
```
when the `type` attribute of the `Constant` node should be "long double".  Note that the declared type is not what matters, it is the parsing of the floating point literal.

eliben:
Good point! Feel free to submit a PR

eliben:
Fixed by #277

jam1garner:
Thanks for properly naming the commits/PR @robbert-harms, it made it much easier to find what caused my compiler to break upon installing on another machine (and therefore getting a new version of pycparser)!

Sparika:
It seems that the case for `long int` is not handled either.

PR #277, opened 2018-08-31
Correct Parsing of Floating Point Literals, issue #253

robbert-harms:
This sets the type attribute to either 'float', 'long double' or 'double' depending on if 'f|F', 'l|L' or '' is specified at the end of the constant definition. This should fix issue #253 .


eliben:
LGTM, thanks!

PR #294, opened 2018-12-01
Fix encoding problem by adding `kwargs` parameter to parse_file function

### LINKED ISSUE #293: parsing some unicode file failed
when parsing some unicode files with function parse_file, the following error occured
`String contains invalid escape code`

After debugging the code, i found `parse_file->preprocess_file->check_output` did not have the `encoding` parameter. if changing the code from file `pycparser\__init__.py:42` to this
` text = check_output(path_list, universal_newlines=True, encoding='utf-8')`
the error disappears

Howerver, i think the correct way to fix this problem is to give users the option to pass `encoding` parameter, and other parameters as well. 
So i suggest adding `**kwargs` to function parse_file and preprocess_file, i am not sure whether i am right or not @eliben

ilovexyz:
Fixes #293 

Description:
For some Chinese Unicode c files, parse_file may fail if not providing proper `encoding` information. Adding `kwargs` parameter will give users the option of providing `encoding` and other parameters necessary for correct parsing.

### DIFF
--- a/pycparser/__init__.py
+++ b/pycparser/__init__.py
@@ -15,7 +15,7 @@ from subprocess import check_output
 from .c_parser import CParser
 
 
-def preprocess_file(filename, cpp_path='cpp', cpp_args=''):
+def preprocess_file(filename, cpp_path='cpp', cpp_args='', **kwargs):
     """ Preprocess a file using cpp.
 
         filename:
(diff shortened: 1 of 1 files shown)


pycparser/__init__.py
```diff
@@ -15,7 +15,7 @@
 from .c_parser import CParser
 
 
-def preprocess_file(filename, cpp_path='cpp', cpp_args=''):
+def preprocess_file(filename, cpp_path='cpp', cpp_args='', **kwargs):
```
eliben: Thinking about it for a bit, this may be premature generalization. For example seeing kwargs in parse_file is unnatural since that function does a number of different things. I think we can just add an "encoding" field with a default value of None for now?

PR #295, opened 2018-12-04
Fix encoding problem by adding `encoding` parameter to parse_file function

### LINKED ISSUE #293: parsing some unicode file failed

ilovexyz:
Fixes #293

Description:
For some Chinese Unicode c files, parse_file may fail if not providing proper `encoding` information. Adding `encoding` parameter will give users the option of providing correct file encoding.

### DIFF
--- a/pycparser/__init__.py
+++ b/pycparser/__init__.py
@@ -15,7 +15,7 @@ from subprocess import check_output
 from .c_parser import CParser
 
 
-def preprocess_file(filename, cpp_path='cpp', cpp_args=''):
+def preprocess_file(filename, cpp_path='cpp', cpp_args='', encoding=None):
     """ Preprocess a file using cpp.
 
         filename:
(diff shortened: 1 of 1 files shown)


eliben:
@ilovexyz I've just reverted this -- please see #296

Issue #299, opened 2018-12-26
Included ply CPP preprocessor does not handle #if a != b correctly 


rreilink:
pycparser.ply.cpp.Preprocessor.evalexpr does an attempt to convern a C-expression to a Python expression by converting ! to not (as in: `!(a==b)`) but inatvertendly it also convert `a!=b` to `a not =b` which is obviously not valid syntax.

Furthermore, undefined identifiers are replaced by `0L` which is not valid Python3 syntax, this should be just `0`.

Additionally, the preprocessing of the evaluation and the actual evaluation should be split into separate methods, such that when deriving from this class, the evaluation could be overridden while the preprocessing (macro expansion etc) stays as-is (in case e.g. due to security concerns one might not want to use eval)

I will prepare a PR.

eliben:
The ply/ dir is just a fork (vendoring) of the https://github.com/dabeaz/ply project, so all issues with its contents should be raised with the project.

Besides, pycparser itself doesn't use ply's cpp.py for anything.

rreilink:
> The ply/ dir is just a fork (vendoring) of the https://github.com/dabeaz/ply project, so all issues with its contents should be raised with the project.
> 
> Besides, pycparser itself doesn't use ply's cpp.py for anything.
@eliben 

1. If these fixes would be included into the dabeaz/ply project, would you be willing to pull them into pycparser? If not, it won't make sense for me to make the effort to do so.

2. Would you be interested in a PR that *does* allow ply's cpp to be used? (i.e. facilitates using plys cpp with pycparser). I have found it more convenient than using the system's cpp e.g. on Windows where cpp is not always available, or when a different cpp version may result in a different result.

eliben:
I occasionally upgrade the PLY version inside pycparser, usually with official releases of PLY.
Note that PLY's cpp is very incomplete, according to its author - I would suggest checking with them for more details. pycparser tries to remain cpp-agnostic on purpose, in order to avoid all this complexity in the project

Issue #300, opened 2018-12-26
Included ply CPP preprocessor does not read files with non-ascii characters


rreilink:
In pycparser.ply.cpp.Preprocessor.include, files are opened as text using unspecified encoding, which result in using `locale.getpreferredencoding(False)`. Using a locale-dependent setting to parse source code is not a good thing. I think a safe default would be to use utf-8 with surrogate escape error handling, which allows all encodings to be decoded without data loss. GCC also uses utf-8 by default.

To provide more flexibility for the user, the reading of the included file could be done in a separate method, allowing the user of the class to override this specific part in a derived class (e.g. to substitute constant text for system include files)

I will prepare a PR.

eliben:
See answer in #299

PR #301, opened 2018-12-26
ply C-preprocessor: Fixes issues #299, #300

rreilink:
Fixes errors in evaluation of #if a!=b (#299) and inclusion of files with non-ascii characters (#300)

Additionally, the include() and evalexpr() methods are split into a few smaller methods, which enables easier customization by overriding specific methods in derived classes


eliben:
See answer in #299

PR #309, opened 2019-02-24
C_generator arrayRef subscript fix

soctar:
If the subscript of an ArrayRef is a constant, the c_generator fails (as '+' cannot be used on a string and an int in python.


soctar:
Ah, this didn't need to happen. I was generating AST that had integers as the value field for constants and the parser keeps things as strings.

PR #310, opened 2019-03-05
Fix crash when file starts with a semicolon

simonlindholm:


tests/test_c_parser.py
```diff
@@ -128,7 +128,7 @@ def test_FileAST(self):
         self.assertEqual(len(t2.ext), 0)
 
     def test_empty_toplevel_decl(self):
-        code = 'int foo;;'
+        code = ';int foo;;'
```
eliben: Please add this as a separate test

tests/test_c_parser.py
```diff
@@ -800,7 +800,7 @@ def test_struct_union(self):
     def test_struct_with_extra_semis_inside(self):
         s1 = """
             struct {
-                int a;;
+                ;int a;;
```
eliben: Same here

simonlindholm:
Updated.

Issue #312, opened 2019-03-22
Generating array type


amirgon:
Consider the following:

```python
from pycparser import c_parser, c_ast, c_generator
parser = c_parser.CParser()
gen = c_generator.CGenerator()

ast = parser.parse("int x[17];")
gen.visit(ast.ext[0])
gen.visit(ast.ext[0].type)
```

Now, `gen.visit(ast.ext[0])` returns the expected result: `'int x[17]'`.
**However, `gen.visit(ast.ext[0].type)` returns: `'int17'`**

I would expect `gen.visit(ast.ext[0].type)` to return `'int[17]'`.  
Returning `'int17'` (the concatenation of `'int'` and `'17'`) just doesn't make sense.  

Trying to figure this out, I looked for `dim` on `c_generator.py` but the [only place it appears](https://github.com/eliben/pycparser/blob/992715f12aef69eb1351308f14fb5f1ff972ce57/pycparser/c_generator.py#L385) handles the concatenation of the square brackets as well, so I'm not sure where the `int17` could come from.

Any idea?

eliben:
I believe this is because `ast.ext[0].type` is an `ArrayDecl`, not a `TypeDecl`. Follow the logic in the `_generate_type` method of the C generator.

Patches welcome!

amirgon:
> Patches welcome!

@eliben #313

amirgon:
Thanks.

PR #313, opened 2019-03-23
Fix array type generation (#312)

amirgon:
Also added `dim_quals` handling to `_generate_type`

Exmaple:

```python
>>> ast = parser.parse('int g(const int a[const 20]){}')
>>> gen.visit(ast.ext[0].decl.type.args.params[0])
'const int a[const 20]'
>>> gen.visit(ast.ext[0].decl.type.args.params[0].type)
'int[const 20]'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@ parser.out
 .tox
 utils/z.c
 *.egg-info
-
+*.swp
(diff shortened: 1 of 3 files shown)


amirgon:
> Thanks for the contribution! Could you please add a test?

@eliben Added tests.

PR #315, opened 2019-03-26
Generate pointer types correctly

amirgon:
I just realized a few things:

- Pointer types are not generated correctly.  
Example:
```python
from pycparser import c_parser, c_ast, c_generator
parser = c_parser.CParser()
gen = c_generator.CGenerator()
ast = parser.parse('const int ** const  x;')

>>> gen.visit(ast.ext[0])
'const int ** const x'
>>> gen.visit(ast.ext[0].type)
'const int'
>>> gen.visit(ast.ext[0].type.type)
'const int'
>>> gen.visit(ast.ext[0].type.type.type)
'const int'
>>> 
```

In this PR I fixed it and now it makes more sense:

```python
>>> gen.visit(ast.ext[0])
'const int ** const x'
>>> gen.visit(ast.ext[0].type)
'const int ** const'
>>> gen.visit(ast.ext[0].type.type)
'const int *'
>>> gen.visit(ast.ext[0].type.type.type)
'const int'
```

- There is code duplication between `visit_ArrayDecl`/`visit_TypeDecl` and `_generate_type`. In order to use `_generate_type`, all that is needed is to prevent it from emitting the declname when we only want to generate these decls.  
In this PR I removed this code duplication.

### DIFF
--- a/tests/test_c_generator.py
+++ b/tests/test_c_generator.py
@@ -336,17 +336,22 @@ class TestCtoC(unittest.TestCase):
         self._assert_ctoc_correct('int g(const int a[const 20]){}')
         ast = parse_to_ast('const int a[const 20];')
         generator = c_generator.CGenerator()
-        self.assertEqual(generator.visit(ast.ext[0].type), 'const int [const 20]')
-        self.assertEqual(generator.visit(ast.ext[0].type.type), 'const int')
+        self.assertEqual(generator.visit(ast.ext[0].type),
+                         'const int [const 20]')
+        self.assertEqual(generator.visit(ast.ext[0].type.type),
+                         'const int')
 
     def test_ptr_decl(self):
         src = 'const int ** const  x;'
         self._assert_ctoc_correct(src)
         ast = parse_to_ast(src)
         generator = c_generator.CGenerator()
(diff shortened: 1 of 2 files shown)


tests/test_c_generator.py
```diff
@@ -336,8 +336,17 @@ def test_array_decl(self):
...
-        self.assertEqual(generator.visit(ast.ext[0].type), 'const int[const 20]')
+        self.assertEqual(generator.visit(ast.ext[0].type), 'const int [const 20]')
         self.assertEqual(generator.visit(ast.ext[0].type.type), 'const int')
 
+    def test_ptr_decl(self):
+        src = 'const int ** const  x;'
+        self._assert_ctoc_correct(src)
+        ast = parse_to_ast(src)
+        generator = c_generator.CGenerator()
+        self.assertEqual(generator.visit(ast.ext[0].type),'const int ** const')
```
eliben: Please have a space after the comma in argument lists. Wrap the second argument if it doesn't fit in 80 lines (see examples elsewhere)

amirgon: Personally I think that the strict 80 character limitation is archaic and irrelevant with today's display sizes.  
Anyway... Fixed on 878527b.

PR #323, opened 2019-04-17
A simple example rewrite Assignment to Augmented Assignment by Visitor

dyhe83:
Another example of rewriting AST node by Visitor.


dyhe83:
OK

Issue #324, opened 2019-05-02
Incorrect Parsing of unsigned int and long int


Sparika:
Similar to issue #253, integer constants with `u/U` and `l/L` suffix are not correctly parsed.

```c
int i = 1uLL;
```
is parsed as 
```
Decl: i, [], [], []
  TypeDecl: i, []
    IdentifierType: ['int']
  Constant: int, 1
```
While it should be 
```
Decl: i, [], [], []
  TypeDecl: i, []
    IdentifierType: ['int']
  Constant: unsigned long long int, 1
```

https://gcc.gnu.org/onlinedocs/gcc/Long-Long.html

PR #326, opened 2019-05-02
Fix issue #324: u/l constant integer suffix

Sparika:
The PR fixes the issue #324.
I've followed the changes made in PR #277

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -1766,14 +1766,26 @@ class CParser(PLYParser):
                         | INT_CONST_HEX
                         | INT_CONST_BIN
         """
-        u = ''
-        l = ''
+        uCount = 0
+        lCount = 0
         for x in p[1][-3:]:
             if x in ('l', 'L'):
-                l += 'long '
+                lCount += 1
             elif x in ('u', 'U'):
-                u = 'unsigned '
-        t = u+l+'int'
+                uCount += 1
+        t = ''
+        if uCount > 1:
+             raise GrammarError('Constant cannot have more than one u/U suffix.')
+        elif lCount > 2:
+             raise GrammarError('Constant cannot have more than two l/L suffix.')
+        else:
+            if uCount:
+                t += 'unsigned '
+            if lCount == 1:
+                t += 'long '
+            elif lCount == 2:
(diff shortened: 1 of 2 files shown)


pycparser/c_parser.py
```diff
@@ -1766,8 +1766,16 @@ def p_constant_1(self, p):
                         | INT_CONST_HEX
                         | INT_CONST_BIN
         """
+        u = ''
+        l = ''
+        for x in p[1][-3:]:
+            if x in ('l', 'L'):
+                l += 'long '
+            elif x in ('u', 'U'):
+                u = 'unsigned '
```
eliben: Could there be more than one unsigned? And if there are multiple `u`s should we have multiple `'unsigned'` modifiers?

Sparika: From what I read, there can be only one `u` and up to two `l`. The `u` may be before or after the `l`s but cannot be between them. Both `l` should have the same case, but not necessarily the same case as the `u`.  I just checked with clang which raises 'invalid suffix' errors on all these cases. 

My implementation is more permissive indeed, I went for simplicity in order to avoid many `if` statements and considering the compiler would be checking for validity. Would you prefer something else?

eliben: How about: instead of accumulating the strings in `u` and `l`, have counters. Increment `uCount` every time `u` or `U` appears, and increment `lCount` similarly for the Ls. Then at the end check that each counter is in valid range (0-1 for u, 0-2 for l), and build up the strings from the counters?

I'm generally OK with accepting slightly malformed code, but what bothers me in this case is that pycparser will generate nonsensical types

Sparika: Ok, I will do that and submit a revision.

Sparika: Ok, I handled the case with too many u/U or l/L raising a ValueError. Note however that the parser seems to already catch this error early on:
```
  File "../pycparser/c_parser.py", line 1867, in p_error
    column=self.clex.find_tok_column(p)))
  File "../pycparser/plyparser.py", line 67, in _parse_error
    raise ParseError("%s: %s" % (coord, msg))
ParseError: :1:35: before: UUU
```

So the case should not happen.

Issue #314, opened 2019-03-25
Failure parsing unnamed function parameters with array dim qualifiers


amirgon:
Consider the following:

```python
from pycparser import c_parser, c_ast, c_generator
parser = c_parser.CParser()
gen = c_generator.CGenerator()

ast1 = parser.parse("int g(int i[const 20]);")
ast2 = parser.parse("int g(int[20]);")
ast3 = parser.parse("int g(int[const 20]);")
```

This code will fail on the last line:
```
>>> ast3 = parser.parse("int g(int[const 20]);")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pycparser/c_parser.py", line 152, in parse
    debug=debuglevel)
  File "pycparser/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "pycparser/ply/yacc.py", line 1199, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "pycparser/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "pycparser/c_parser.py", line 1847, in p_error
    column=self.clex.find_tok_column(p)))
  File "pycparser/plyparser.py", line 67, in _parse_error
    raise ParseError("%s: %s" % (coord, msg))
pycparser.plyparser.ParseError: :1:11: before: const
```

eliben:
Thanks for the report. Feel free to submit a PR

PR #327, opened 2019-05-31
Fix issue #314: Failed parsing unnamed function parameters with array dim qualifiers

saullocarvalho:
* Add unnamed function parameter tests
* Fix grammar rule


eliben:
Thanks!

Issue #187, opened 2017-04-12
Feature request: Ability to provide preexisting typedefs to the parser


rhelmot:
Our usecase for pycparser is allowing the user to provide snippets of c variable and function declarations and we convert them to a domain-specific representation of the types used, useful for further analysis. This, as I understand, breaks an assumption baked into pycparser that it's parsing files which can be compiled, so there are complications. What we'd like to be able to do is allow the user of our tool to parse some typedefs and some variable declarations using those types separately. This means that right now, we have to prepend every string that we feed to `parse()` with all the prior typedefs, which is awkward since we don't have very good support for serializing our internal representation of the types back to strings.

What we'd like to be able to do is to provide a mapping, either to the `CParser` constructor or to `parse()`, of pre-established typedefs that it may use, either as strings or as ASTs, whatever is the most convenient.

rhelmot:
Additionally, the ability to just declare new type names without describing a C type they resolve to would be neat, so we can parse types which may be typedef'd differently on different systems like `uint32_t` agnostically

eliben:
It's not clear to me if what you're asking is simply a convenience API/facade moved into pycparser rather than living outside it. How do you expect to provide preexisting typedefs to pycparser in the proposed approach? Can you sketch some how-it-might-look code?

rhelmot:
Sure. Something along these lines:

```python
>>> import pycparser
>>> p = pycparser.Parser()
>>> p.add_predef_type('foo', ['struct', 'foo'])
>>> p.add_predef_type('bar', ['long', 'int'])
>>> p.parse('foo x;').children()[0][-1].type.type
<pycparser.c_ast.Struct at 0x7f3ae5000780>
>>> p.parse('foo x;').children()[0][-1].type.type.name
'foo'
>>> p.parse('bar x;').children()[0][-1].type.type
<pycparser.c_ast.IdentifierType at 0x7f3ae471b830>
>>> p.parse('bar x;').children()[0][-1].type.type.names
['long', 'int']
```

For the second part, I don't know how feasible this would be, but:

```python
>>> p.add_base_type('baz')
>>> p.parse('baz x;').children()[0][-1].type.type
<pycparser.c_ast.IdentifierType at 0x7f3ae4717050>
>>> p.parse('baz x;').children()[0][-1].type.type.names
['baz']
```

eliben:
I guess I don't understand what the big advantage of this approach is vs. just prepending a `typedef` to the parsed code.

[...]

PR #328, opened 2019-06-03
Fix issue #187: Ability to provide preexisting typedefs to the parser

saullocarvalho:


eliben:
Thanks, but my comments in #187 still apply. I don't believe this feature is necessary in the core pycparser. It can be easily added in user programs, so I prefer to keep the core simple

Issue #99, opened 2015-09-25
parser for FuncDecl incorrectly sets declname attribute on return type


overdamped:
Steps to reproduce in the examples folder running python interpreter.

```
>>> from pycparser import parse_file, c_generator
>>> from pycparser.c_ast import *
>>> ast = parse_file("c_files/memmgr.h", use_cpp=True)
>>> cg = c_generator.CGenerator()
>>> cg.visit(Cast(ast.ext[-3].type.type, FuncCall(ID('test_fun'), ExprList([]))))
'(void *memmgr_alloc) test_fun()'
>>> #workaround
... ast.ext[-3].type.type.type.declname = ''
>>> cg.visit(Cast(ast.ext[-3].type.type, FuncCall(ID('test_fun'), ExprList([]))))
'(void *) test_fun()'

```

eliben:
Thanks for the report. Would you like to try your hand at fixing this?

overdamped:
I'll try and dig at it a bit this week, but no promises. Thanks for quick response.

Juzley:
Is this definitely a bug? Generally TypeDecls associated with declarators have the identifier's name associated with them in the declname, and the CGenerator uses that. I think the correct way to achieve what you're trying to do is to create a new type for your cast, which doesn't have an associated declname, something like:

`
func_type = ast.ext[-3].type.type
cast_type = PtrDecl(func_type.quals, TypeDecl(None, func_type.type.quals, func_type.type.type))
cg.visit(Cast(cast_type, FuncCall(ID('test_fun'), ExprList([]))))
`

This gives the expected output:
`(void *) test_fun()`

PR #329, opened 2019-06-07
Fix issue #99: parser for FuncDecl incorrectly sets declname attribute on return type

saullocarvalho:
visit_Cast was emitting declname (default behavior) for to_type attribute incorrectly.

### DIFF
--- a/tests/test_c_generator.py
+++ b/tests/test_c_generator.py
@@ -1,15 +1,12 @@
 import os
 import platform
 import sys
-import textwrap
 import unittest
 
 # Run from the root dir
 sys.path.insert(0, '.')
 
 from pycparser import c_parser, c_generator, c_ast, parse_file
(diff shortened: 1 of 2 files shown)


tests/test_c_generator.py
```diff
@@ -1,11 +1,15 @@
...
 import textwrap
 import unittest
 
 # Run from the root dir
 sys.path.insert(0, '.')
 
-from pycparser import c_parser, c_generator, c_ast
+from pycparser import c_parser, c_generator, c_ast, parse_file
+
+CPPPATH = 'cpp'
```
eliben: I prefer not to use `cpp` in these tests and have chunks of preprocessed code instead

saullocarvalho: I've removed cpp!

PR #346, opened 2019-08-21
Fix error transforming an empty switch

### LINKED ISSUE #345: "'NoneType' object is not iterable" seen for switch statement without cases
Seen calling parse_file on a file containing the below snippet (similar code was generated by a preprocessor). This compiles with `gcc`. I think checking if block_items is None may be sufficient.

```c
int main() {
    int type = 2;
    switch(type) {}
}
```

```
/usr/local/lib/python3.7/site-packages/pycparser/ply/yacc.py in parseopt_notrack(self, input, lexer, debug, tracking, tokenfunc)
   1116                             del symstack[-plen:]
   1117                             self.state = state
-> 1118                             p.callable(pslice)
   1119                             del statestack[-plen:]
   1120                             symstack.append(sym)

/usr/local/lib/python3.7/site-packages/pycparser/c_parser.py in p_selection_statement_3(self, p)
   1505         """ selection_statement : SWITCH LPAREN expression RPAREN pragmacomp_or_statement """
   1506         p[0] = fix_switch_cases(
-> 1507                 c_ast.Switch(p[3], p[5], self._token_coord(p, 1)))
   1508 
   1509     def p_iteration_statement_1(self, p):

/usr/local/lib/python3.7/site-packages/pycparser/ast_transforms.py in fix_switch_cases(switch_node)
     75     # Goes over the children of the Compound below the Switch, adding them
     76     # either directly below new_compound or below the last Case as appropriate
---> 77     for child in switch_node.stmt.block_items:
     78         if isinstance(child, (c_ast.Case, c_ast.Default)):
     79             # If it's a Case/Default:

TypeError: 'NoneType' object is not iterable
```

TysonAndre:
The parser would throw on that line for `switch(1) {}`
because NoneType is not iterable.

Fixes #345

### DIFF
--- a/tests/test_c_parser.py
+++ b/tests/test_c_parser.py
@@ -1792,6 +1792,7 @@ class TestCParser_whole_code(TestCParser_base):
         switch = ps1.ext[0].body.block_items[0]
 
         block = switch.stmt.block_items
+        self.assertEqual(len(block), 4)
         assert_case_node(block[0], '10')
         self.assertEqual(len(block[0].stmts), 3)
         assert_case_node(block[1], '20')
@@ -1799,7 +1800,6 @@ class TestCParser_whole_code(TestCParser_base):
         assert_case_node(block[2], '30')
         self.assertEqual(len(block[2].stmts), 1)
         assert_default_node(block[3])
-        self.assertEqual(len(block), 4)
 
         s2 = r'''
         int foo(void) {
@@ -1820,6 +1820,7 @@ class TestCParser_whole_code(TestCParser_base):
         switch = ps2.ext[0].body.block_items[0]
 
         block = switch.stmt.block_items
+        self.assertEqual(len(block), 5)
         assert_default_node(block[0])
         self.assertEqual(len(block[0].stmts), 2)
         assert_case_node(block[1], '10')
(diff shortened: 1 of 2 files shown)


TysonAndre:
Done, asserted that the new matches were `[]` in test_switch_statement. I couldn't find that test the first time I looked.

tests/test_c_parser.py
```diff
@@ -1799,6 +1799,7 @@ def assert_default_node(node):
         assert_case_node(block[2], '30')
         self.assertEqual(len(block[2].stmts), 1)
         assert_default_node(block[3])
+        self.assertEqual(len(block), 4)
```
eliben: Could you please move this test to before the assertions on `block[...]` begin?

tests/test_c_parser.py
```diff
@@ -1829,6 +1830,20 @@ def assert_default_node(node):
...
+            switch (myvar) {
+            }
+            return 0;
+        }
+        '''
+        ps3 = self.parse(s3)
+        switch = ps3.ext[0].body.block_items[0]
+
+        block = switch.stmt.block_items
+        self.assertEqual(block, [])
```
eliben: Nit: `block` is unnecessary here, just assert on `switch.stmt.block_items`

PR #347, opened 2019-08-21
Fix slow backtracking when parsing strings (no external deps)

### LINKED ISSUE #61: String literals with many escapes take a long time to tokenize
Running the following program on the following data takes a long time to parse.

``` python
if __name__ == '__main__':
    ast = pycparser.parse_file("x.c")
    ast.show()
```

``` c
int main(int argc, char** argv) {
    "\123\123\123\123\123\123\123\123\123\123\123\123\123\123\123";
}
```

I think the problem is that the BAD_STRING_LITERAL regular expression is doing backtracking, resulting in exponential running time. Add another '\123' to the string and it will take roughly 3 times as long.

If I break the BAD_STRING_LITERAL pattern, by requiring that it start with some unexpected string `'xyzzy and then some'` so that it never starts to match, the program runs quickly.

TysonAndre:
Fixes #61

This uses negative lookaheads to avoid ambiguity in how string should be
parsed by the regex.
- https://docs.python.org/2/library/re.html#regular-expression-syntax

- Previously, if it didn't immediately succeed at parsing an escape
  sequence such as `\123`, it would have to try `\1`+`23`, `\12` + `3`,
  and `\123`, which multiplied the time taken by 3 per additional escape
  sequence. This solves that by only allowing `\123`
- The same fix was added for hex escapes.

Also fix a test that relied on the incorrect handling of regexes.
The implementation documentation says that it intends to allow
**decimal** escapes permissively.

### DIFF
--- a/pycparser/c_lexer.py
+++ b/pycparser/c_lexer.py
@@ -206,9 +206,9 @@ class CLexer(object):
     # code.
     #
     simple_escape = r"""([a-zA-Z._~!=&\^\-\\?'"])"""
-    decimal_escape = r"""(\d+)"""
-    hex_escape = r"""(x[0-9a-fA-F]+)"""
-    bad_escape = r"""([\\][^a-zA-Z._~^!=&\^\-\\?'"x0-7])"""
+    decimal_escape = r"""(\d+)(?!\d)"""
+    hex_escape = r"""(x[0-9a-fA-F]+)(?![0-9a-fA-F])"""
+    bad_escape = r"""([\\][^a-zA-Z._~^!=&\^\-\\?'"x0-9])"""
 
     escape_sequence = r"""(\\("""+simple_escape+'|'+decimal_escape+'|'+hex_escape+'))'
     cconst_char = r"""([^'\\\n]|"""+escape_sequence+')'
--- a/tests/test_c_lexer.py
+++ b/tests/test_c_lexer.py
@@ -116,6 +116,7 @@ class TestCLexerNoErrors(unittest.TestCase):
         self.assertTokensTypes(r"""'\t'""", ['CHAR_CONST'])
         self.assertTokensTypes(r"""'\''""", ['CHAR_CONST'])
         self.assertTokensTypes(r"""'\?'""", ['CHAR_CONST'])
+        self.assertTokensTypes(r"""'\0'""", ['CHAR_CONST'])
         self.assertTokensTypes(r"""'\012'""", ['CHAR_CONST'])
         self.assertTokensTypes(r"""'\x2f'""", ['CHAR_CONST'])
         self.assertTokensTypes(r"""'\x2f12'""", ['CHAR_CONST'])
@@ -148,6 +149,13 @@ class TestCLexerNoErrors(unittest.TestCase):
             ['STRING_LITERAL'])
         self.assertTokensTypes(
             '"\123\123\123\123\123\123\123\123\123\123\123\123\123\123\123\123"',
+            ['STRING_LITERAL'])
+        # The lexer is permissive and allows decimal escapes (not just octal)
+        self.assertTokensTypes(
+            '"jx\9"',
+            ['STRING_LITERAL'])
+        self.assertTokensTypes(
+            '"fo\9999999"',
             ['STRING_LITERAL'])
 
     def test_mess(self):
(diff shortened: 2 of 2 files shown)


TysonAndre:
This still has backtracking for `"\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\`\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x23\x2\x23456` without a closing quote. Investigating.

TysonAndre:
In the previous commits,
`\x12` could be parsed both as `\x`+`12` and `\x12`,
which caused exponential options for backtracking, because "\a-z" was allowed for windows paths.

I solved this by allowing "\x" when not followed by hex
in the regular string literal (and treating \x followed by at least one hex character as hex)

pycparser/c_lexer.py
```diff
@@ -205,10 +205,10 @@ def _make_tok_location(self, token):
     # parse all correct code, even if it means to sometimes parse incorrect
     # code.
     #
-    simple_escape = r"""([a-zA-Z._~!=&\^\-\\?'"])"""
-    decimal_escape = r"""(\d+)"""
-    hex_escape = r"""(x[0-9a-fA-F]+)"""
-    bad_escape = r"""([\\][^a-zA-Z._~^!=&\^\-\\?'"x0-7])"""
+    simple_escape = r"""([a-wyzA-Z._~!=&\^\-\\?'"]|x(?![0-9a-fA-F]))"""
```
eliben: Please add a sizable comment explaining how these regexes are structured and why. The original regexes were taken verbatim from the C syntax definition, IIRC

pycparser/ply/lex.py
```diff
@@ -317,7 +317,13 @@ def token(self):
 
             # Look for a regular expression match
             for lexre, lexindexfunc in self.lexre:
-                m = lexre.match(lexdata, lexpos)
+                try:
```
eliben: Debugging leftovers?

TysonAndre: Yes, they are. Removed.

TysonAndre:
> Thanks for pursuing this. I'm thinking about the added complexity vs. performance win in edge cases. It's not clear to me it's worth it, yet, to be honest. One of the things I'm considering is whether this reduces performance in more "regular" use cases, but unfortunately pycparser doesn't have a benchmark suite.

Added another simplified regex just for strings (but not character literals, which expect just one escape sequence). The unit tests I added continue to pass.

This takes advantage of the fact that `\[0-9]` is valid on its own, and `\x` is valid on its own, due to the permissive nature, and we don't need the following sequences of non-escaped characters themselves.

eliben:
Could you please run the benchmark script I added in https://github.com/eliben/pycparser/tree/master/utils/benchmark without and then with your changes applied and report the results here?

TysonAndre:
1. In general, benchmarking results on my laptop have been all over the place for various projects. I'd assumed this was due to mitigations for intel prefetching bugs, power saving, etc., but I'm not actually sure
2. The script fails when run on utils/benchmark/inputs due to a ParseError on .gitattributes. I removed that to run the script

Example results are included below, for origin/master and my branch

```
±3b89659⚡ » taskset 1 python3 utils/benchmark/benchmark-parse.py utils/benchmark/inputs 
tccgen.c.ppout           .....    Mean: 0.860  Stddev: 0.014
sqlite-btree.c.ppout     .....    Mean: 1.385  Stddev: 0.012
redis.c.ppout            .....    Mean: 0.586  Stddev: 0.040
±3b89659⚡ » taskset 1 python3 utils/benchmark/benchmark-parse.py utils/benchmark/inputs 
tccgen.c.ppout           .....    Mean: 1.061  Stddev: 0.031
sqlite-btree.c.ppout     .....    Mean: 1.502  Stddev: 0.028
redis.c.ppout            .....    Mean: 0.527  Stddev: 0.016
±3b89659⚡ » taskset 1 python3 utils/benchmark/benchmark-parse.py utils/benchmark/inputs 
tccgen.c.ppout           .....    Mean: 0.873  Stddev: 0.012
sqlite-btree.c.ppout     .....    Mean: 1.407  Stddev: 0.011
redis.c.ppout            .....    Mean: 0.604  Stddev: 0.032
```

```
±regex-fix-backtracking-v2⚡ » taskset 1 python3 utils/benchmark/benchmark-parse.py utils/benchmark/inputs 
tccgen.c.ppout           .....    Mean: 0.952  Stddev: 0.133
sqlite-btree.c.ppout     .....    Mean: 1.383  Stddev: 0.004
redis.c.ppout            .....    Mean: 0.543  Stddev: 0.052
±regex-fix-backtracking-v2⚡ » taskset 1 python3 utils/benchmark/benchmark-parse.py utils/benchmark/inputs 
tccgen.c.ppout           .....    Mean: 0.868  Stddev: 0.014
sqlite-btree.c.ppout     .....    Mean: 1.389  Stddev: 0.005
redis.c.ppout            .....    Mean: 0.531  Stddev: 0.055
```

Issue #224, opened 2017-12-04
fake_libc_include not included in distributed version of library


inglesp:
The `fake_libc_include` directory is not included in the version of the library installed from PyPI with pip.

This means that I cannot do something like:

```
pycparser_path = os.path.basename(pycparser.__file__)
include_path = os.path.join(pycparser_path, 'utils', 'fake_libc_include')
ast = parse_file(filename, use_cpp=True, cpp_args=rf'-I{include_path}')
```

Instead I have to vendor `fake_libc_include` (which is not the end of the world!).

Would you consider a patch to fix this?

inglesp:
s/consider/consider accepting/

eliben:
It's listed here: https://github.com/eliben/pycparser/blob/master/MANIFEST.in#L4

And I do see it in the distribution tarball -- are you sure pip doesn't install it?

inglesp:
I'm pretty sure... I've `pip install`ed the libary into a fresh virtualenv, and it's not there:

```
~/.pyenv/versions/pycparser-test/lib/python3.6/site-packages/pycparser $ tree
.
├── __init__.py
├── __pycache__
│   ├── __init__.cpython-36.pyc
│   ├── _ast_gen.cpython-36.pyc
│   ├── _build_tables.cpython-36.pyc
│   ├── ast_transforms.cpython-36.pyc
│   ├── c_ast.cpython-36.pyc
│   ├── c_generator.cpython-36.pyc
│   ├── c_lexer.cpython-36.pyc
│   ├── c_parser.cpython-36.pyc
│   ├── lextab.cpython-36.pyc
│   ├── plyparser.cpython-36.pyc
│   └── yacctab.cpython-36.pyc
├── _ast_gen.py
├── _build_tables.py
├── _c_ast.cfg
├── ast_transforms.py
├── c_ast.py
├── c_generator.py
├── c_lexer.py
├── c_parser.py
├── lextab.py
├── ply
│   ├── __init__.py
│   ├── __pycache__
│   │   ├── __init__.cpython-36.pyc
│   │   ├── cpp.cpython-36.pyc
│   │   ├── ctokens.cpython-36.pyc
│   │   ├── lex.cpython-36.pyc
│   │   ├── yacc.cpython-36.pyc
│   │   └── ygen.cpython-36.pyc
│   ├── cpp.py
│   ├── ctokens.py
│   ├── lex.py
│   ├── yacc.py
│   └── ygen.py
├── plyparser.py
[...]

PR #358, opened 2019-12-08
Update README to warn about missing support files

WasabiFan:
#224


WasabiFan:
I'll let you pick which place you'd prefer (you can edit this PR from the web interface). I think the fact they're included in the tarball doesn't really matter because it won't get installed. As long as this is clarified so people don't waste time on it, I'm happy.

PR #364, opened 2020-02-25
Fix #363 incorrect AST when parsing offsetof

### LINKED ISSUE #363: Incorrect AST structure when parsing `offsetof` expressions.
I am using the latest commit on `master` branch (`1166ea1`).

Take the following code as example:
```c
int x = offsetof(struct A, a.b);
```
The generated parse tree is:
```python
FileAST(ext=[Decl(name='x',
                  quals=[],
                  storage=[],
                  funcspec=[],
                  type=TypeDecl(declname='x',
                                quals=[],
                                type=IdentifierType(names=['int'])),
                  init=FuncCall(name=ID(name='offsetof'),
                                args=ExprList(exprs=[Typename(name=None,
                                                              quals=[],
                                                              type=TypeDecl(declname=None,
                                                                            quals=[],
                                                                            type=Struct(name='A',
                                                                                        decls=None))),
                                                     StructRef(name=ID(name='a'),
                                                               type='.',
                                                               field=ID(name=ID(name='b')))])),
                  bitsize=None)])
```
where the `field` attribute of `StructRef` has a value of `ID(name=ID(name='b'))` instead of just `ID(name='b')`.

I think the issue is with the production rules of `offsetof_member_designator` that was introduced in #145, namely:
https://github.com/eliben/pycparser/blob/1166ea11785ce12cdfd5e8bf8b3a69b5e6b76f9c/pycparser/c_parser.py#L1735-L1748
In the `len(p) == 4` case, `StructRef` should be created using `p[3]` directly, instead of constructing an `ID` first.

If you agree with my analysis, I'd be happy to create a patch.

huzecong:
This PR addresses #363, where parsing `offsetof` expressions generates incorrect ASTs.

The main change is eliminating the duplicate `ID` node constructed in `p_offsetof_member_designator`. New tests are added to check the entire structure of parsed `StructRef` and `ArrayRef` nodes in an `offsetof` expression.

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -1740,8 +1740,7 @@ class CParser(PLYParser):
         if len(p) == 2:
             p[0] = p[1]
         elif len(p) == 4:
-            field = c_ast.ID(p[3], self._token_coord(p, 3))
-            p[0] = c_ast.StructRef(p[1], p[2], field, p[1].coord)
+            p[0] = c_ast.StructRef(p[1], p[2], p[3], p[1].coord)
         elif len(p) == 5:
             p[0] = c_ast.ArrayRef(p[1], p[3], p[1].coord)
         else:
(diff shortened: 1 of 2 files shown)


eliben:
Thanks for the contribution, and apologies for the delay. I've been very busy lately, and expect to have some time for pycparser in the next few weeks.

PR #368, opened 2020-03-08
Add some "types" to zlib.h

thoni56:
Since `pycparser` includes a `zlib.h` some of the types defined by the real `zlib.h` is also required.

I'm not sure if the guarding #ifdef should be around the inclusion of the other fakes.

### DIFF
--- a/utils/fake_libc_include/zlib.h
+++ b/utils/fake_libc_include/zlib.h
@@ -1,2 +1,21 @@
 #include "_fake_defines.h"
 #include "_fake_typedefs.h"
+
+#define uInt int
+#define uLong int
+#define Byte int
+
+/* Some FAR types */
+#define Bytef int
+#define charf int
+#define inft inf
+
+#define voidpc int
+#define voidpf int
+#define voidp int
+
+#define z_crc_t int
+#define z_size_t int
+
+#define alloc_func int
+#define free_func int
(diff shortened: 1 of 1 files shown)


thoni56:
Fixed the review comments, and also tried to mimic what is conditionally defined in zlib's `zconf.h`.

utils/fake_libc_include/zlib.h
```diff
@@ -1,2 +1,33 @@
...
+
+typedef int uInt
+typedef int uLong
+#if !defined(__MACTYPES__)
+typedef int Byte
+#endif
+
+typedef int Bytef
+typedef int charf
+typedef inf intf
```
eliben: Is this really `inf`?

thoni56: Arrgh. Sorry about that, I must have been asleep when doing this. You missed all the missing semi-colons ;-)

My test runs included the standard `fake_lib` instead, which masked the updated one. With https://github.com/eliben/pycparser/pull/368/commits/9557a944b4bac002b6a1fb6532c79a25ba715a5e it compiles with `gcc` 7.4, except for a duplication of `__built_in_va_list` which is not related to `zlib.h`:

```
e$ gcc zlib.h
In file included from zlib.h:5:0:
_fake_typedefs.h:5:13: error: conflicting types for ‘__builtin_va_list’
 typedef int __builtin_va_list;
             ^~~~~~~~~~~~~~~~~
cc1: note: previous declaration of ‘__builtin_va_list’ was here
```

Issue #378, opened 2020-05-26
Replace assertions with checks that raise `ParseError`s


huzecong:
https://github.com/eliben/pycparser/blob/61eac63fda12f9ba4504def70d0d76cd3ac1bbff/pycparser/c_parser.py#L464-L467

In this and other lines of `c_parser.py`, assertions are used to verify that said condition is true. However, there can be input that will result in the condition being false. For example, the following code triggers an `AssertionError` at the line above:
```c
typedef struct __attribute__((packed)) {
    char name[36];
    int attr;
} foo;
```
I would argue that the desired behavior here is to be consistent and throw a `ParseError` instead.

eliben:
Feel free to send PRs for concrete cases. Let's start with one and see how it goes from there

huzecong:
A PR for this issue sounds simple enough; I'll try to create one tomorrow.

Issue #385, opened 2020-07-11
Exponential time complexity when generating code from AST with nested `sizeof`s.


huzecong:
Minimum example to reproduce:
```python
from pycparser.c_parser import CParser
from pycparser.c_generator import CGenerator

# Code is 30 nested `sizeof`s.
code = "int x = sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(sizeof(int))))))))))))))))))))))))))))));"
tree = CParser().parse(code)
print(tree)  # the tree is correct
print(CGenerator().visit(tree))  # this never stops
```

The issue is within `CGenerator.visit_UnaryOp`:
https://github.com/eliben/pycparser/blob/61eac63fda12f9ba4504def70d0d76cd3ac1bbff/pycparser/c_generator.py#L61-L72
In the case where `n.op` is `sizeof`, `n.expr` is visited twice. When there are nested `sizeof`s, the leaf nodes will be visited an exponential number of times. (Although in reality such code wouldn't make sense)

I can create a PR with a few of my previous issues, but it might take a few days before I can work on this.

eliben:
Interesting find. Feel free to send a PR to fix this.

Issue #379, opened 2020-05-27
`AttributeError` when parsing `struct` with nested `enum`


huzecong:
When parsing the following code:
```c
typedef struct {
  enum {
    A = 1
  };
} foo;
```
The following exception is raised:
```python
~/.pyenv/versions/3.7.2/Python.framework/Versions/3.7/lib/python3.7/site-packages/pycparser/c_parser.py in parse(self, text, filename, debuglevel)
    150                 input=text,
    151                 lexer=self.clex,
--> 152                 debug=debuglevel)
    153
    154     ######################--   PRIVATE   --######################

~/.pyenv/versions/3.7.2/Python.framework/Versions/3.7/lib/python3.7/site-packages/pycparser/ply/yacc.py in parse(self, input, lexer, debug, tracking, tokenfunc)
    329             return self.parseopt(input, lexer, debug, tracking, tokenfunc)
    330         else:
--> 331             return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
    332
    333

~/.pyenv/versions/3.7.2/Python.framework/Versions/3.7/lib/python3.7/site-packages/pycparser/ply/yacc.py in parseopt_notrack(self, input, lexer, debug, tracking, tokenfunc)
   1116                             del symstack[-plen:]
   1117                             self.state = state
-> 1118                             p.callable(pslice)
   1119                             del statestack[-plen:]
   1120                             symstack.append(sym)

~/.pyenv/versions/3.7.2/Python.framework/Versions/3.7/lib/python3.7/site-packages/pycparser/c_parser.py in p_struct_declaration_1(self, p)
    993             decls = self._build_declarations(
    994                 spec=spec,
--> 995                 decls=[dict(decl=decl_type)])
    996
    997         else:

~/.pyenv/versions/3.7.2/Python.framework/Versions/3.7/lib/python3.7/site-packages/pycparser/c_parser.py in _build_declarations(self, spec, decls, typedef_namespace)
    418             decls_0_tail = decls[0]['decl']
    419             while not isinstance(decls_0_tail, c_ast.TypeDecl):
--> 420                 decls_0_tail = decls_0_tail.type
    421             if decls_0_tail.declname is None:
    422                 decls_0_tail.declname = spec['type'][-1].names[0]

AttributeError: 'Enum' object has no attribute 'type'
```

Not sure if the code is standard-conforming, but in any case an `AttributeError` should not be thrown.

eliben:
[...]

PR #387, opened 2020-07-17
Fix issues #378, #379, #385

huzecong:
This PR address the following issues:

- #378: Replace assertion inside `CParser` with check that raises `ParseError` on failure.
  **What's changed:** Only the assertion inside `_build_function_definition` is replaced. The assertion is not appropriate because there are possible inputs (although grammatically incorrect) that would trigger the assertion.
- #379: Fix exception when parsing `struct` with nested `enum`.
  **What's changed:** `Enum` is also added to the list of node types that are handled differently inside `CParser._build_declarations` (the others being `Struct`, `Union`, and `IdentifierType`).
- #385: Fix exponential time cost when generating code with nested `sizeof`s.
  **What's changed:** Simple fix in `CGenerator`.

New test cases are added to cover the changes.

### DIFF
--- a/pycparser/c_generator.py
+++ b/pycparser/c_generator.py
@@ -59,17 +59,18 @@ class CGenerator(object):
         return fref + '(' + self.visit(n.args) + ')'
 
     def visit_UnaryOp(self, n):
-        operand = self._parenthesize_unless_simple(n.expr)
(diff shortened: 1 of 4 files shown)


huzecong:
Got it, will do that in the future!

PR #394, opened 2020-10-01
Added flattening of abundant parenthesis in CGenerator

### LINKED ISSUE #393: Precedence-aware CGenerator
I ran into the issue that in a code more than 256 variables are summed up. The `CGenerator` turned this into >256 nested BinaryOperators with that many parenthesis, that in turn was something clang did not like (be default).

I wrote a small generator which takes precedence into account for binary operators:

```python
class LessParanthesizingCGenerator(CGenerator):
    def get_BinaryOp_precedence(self, n):
        """
        Gives precedence for op of n, otherwise -1.
        
        Lower number have precedence over higher numbers.
        """
        binary_op_precedence = {
            # based on https://en.cppreference.com/w/c/language/operator_precedence
            '*': 3, '%': 3,
            '-': 4, '+': 4,
            '<<': 5, '>>': 5,
            '<': 6, '<=': 6, '>': 6, '>=': 6,
            '==': 7, '!=': 7,
            '&': 8,
            '^': 9,
            '|': 10,
            '&&': 11,
            '||': 12
        }
        if not isinstance(n, c_ast.BinaryOp):
            return -1
        else:
            return binary_op_precedence[n.op]

    def visit_BinaryOp(self, n):
        p = self.get_BinaryOp_precedence(n)
        lval_str = self._parenthesize_if(n.left,
                            lambda d: not self._is_simple_node(d) and 
                                      self.get_BinaryOp_precedence(d) > p)
        rval_str = self._parenthesize_if(n.right,
                            lambda d: not self._is_simple_node(d) and
                                      self.get_BinaryOp_precedence(d) >= p)
        return '%s %s %s' % (lval_str, n.op, rval_str)
```

Would you like me to make a pull request and if so, is this implementation sufficient for now? Tests I would of cause add.

cod3monk:
Use with `CGenerator(flatten=True)`:
```python
>>> from pycparser import CParser
>>> from pycparser.c_generator import CGenerator
>>>
>>> ast = CParser().parse('int x = a+b+c*d;')
__main__:264: DeprecationWarning: 'U' mode is deprecated
>>> CGenerator().visit(ast)
'int x = (a + b) + (c * d);\n'
>>> CGenerator(flatten=True).visit(ast)
'int x = a + b + c * d;\n'
>>>
```

Fixes #393.


pycparser/ply/ctokens.py
```diff
@@ -12,7 +12,7 @@
     'ID', 'TYPEID', 'INTEGER', 'FLOAT', 'STRING', 'CHARACTER',
 
     # Operators (+,-,*,/,%,|,&,~,^,<<,>>, ||, &&, !, <, <=, >, >=, ==, !=)
-    'PLUS', 'MINUS', 'TIMES', 'DIVIDE', 'MODULO',
+    'PLUS', 'MINUS', 'TIMES', 'DIVIDE', 'MOD',
```
eliben, line 15: Can't change this file. `ply` is a vendored dependency

pycparser/c_generator.py
```diff
@@ -72,11 +73,54 @@ def visit_UnaryOp(self, n):
...
+        'NE': '!='
+    }
+
+    @property
+    def _prededence_map(self):
+        """
+        Map of operands (e.g., '%') to precedence (higher number -> stronger precedence).
+        """
+        if not hasattr(self, '_prededence_map_cache'):
+            self._prededence_map_cache = {}
```
eliben, line 105: I may have misled you here earlier, sorry. I'd actually prefer this not to depend on `c_parser` -- this seems like a needless dependency that can complicate things.

I see two options: either we leave this just duplicated in two places - it's not like the C operator precedence is going to change! Or we refactor it to a separate file which both `c_parser` and `c_generator` import. WDYT?

pycparser/c_generator.py
```diff
@@ -6,19 +6,20 @@
...
+from . import c_ast, c_parser
 
 
 class CGenerator(object):
     """ Uses the same visitor pattern as c_ast.NodeVisitor, but modified to
         return a value from each visit method, using string accumulation in
         generic_visit.
     """
-    def __init__(self):
+    def __init__(self, flatten=False):
```
eliben, line 17: I'd suggest a more descriptive name for this option. `flatten` can mean meany things.

Something like `minimal_parens`, etc

pycparser/c_generator.py
cod3monk: I duplicated it and added a comment do `c_parser.py`

pycparser/c_generator.py
cod3monk: Using `reduce_parentheses`

cod3monk:
Sorry for being a little trigger-happy on commits ;)

pycparser/c_generator.py
```diff
@@ -14,11 +14,12 @@ class CGenerator(object):
         return a value from each visit method, using string accumulation in
         generic_visit.
     """
-    def __init__(self):
+    def __init__(self, reduce_parentheses=False):
         # Statements start with indentation of self.indent_level spaces, using
```
eliben, line 18: The comment should explain the new parameter and what it means

pycparser/c_generator.py
```diff
@@ -72,11 +73,31 @@ def visit_UnaryOp(self, n):
             else:
                 return '%s%s' % (n.op, operand)
 
+    precedence_map = {
+        # Some what of a duplicate of c_generator.CGenerator.precedence
```
eliben, line 77: Did you mean to refer to c_parser here?

pycparser/c_generator.py
```diff
@@ -72,11 +73,31 @@ def visit_UnaryOp(self, n):
...
+    }
+
     def visit_BinaryOp(self, n):
-        lval_str = self._parenthesize_if(n.left,
-                            lambda d: not self._is_simple_node(d))
-        rval_str = self._parenthesize_if(n.right,
-                            lambda d: not self._is_simple_node(d))
+        lval_str = self._parenthesize_if(
+            n.left,
+            lambda d: not (self._is_simple_node(d) or
```
eliben, line 93: Add a comment explaining the logic here; why you use `>=` for lhs and `>` for rhs, etc.

eliben (changes requested):

eliben:
> Sorry for being a little trigger-happy on commits ;)

No worries, I squash them anyway.

When you're done addressing the code review, please either "Request code review " again (using the refresh icon next to my name), or add a comment saying so

pycparser/c_generator.py
cod3monk: yes :)

eliben (approved):

PR #398, opened 2020-12-06
Add testing and support for modern Pythons

jdufresne:
Python 3.9 was released on October 5th, 2020.


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py34,py35,py36
+envlist = py27,py34,py35,py36,py37,py38,py39
```
eliben, line 2: I'd rather not touch tox.ini at this point. I'm looking at getting rid of Tox entirely

eliben (changes requested):

tox.ini
jdufresne: No problem, done.

Just curious, why? `tox` is quite convenient for testing across many python versions locally. NBD, just curious.

jdufresne: And, if that's the case, would you like me to open a PR to delete it?

eliben: Just removed it, to make things simpler.

For multi-version testing I actually find GitHub actions more useful since it gives me access to more versions without having to install and juggle them locally.

eliben (approved):

PR #400, opened 2020-12-06
Simplify test execution

jdufresne:
Use the stdlib standard entry point for running tests through the
command:

    python -m unittest discover

Docs: https://docs.python.org/3/library/unittest.html#unittest-test-discovery

This automatically looks for files with the test_ prefix and runs them
as tests. This removes the need for the custom test entry point script,
all_tests.py.


eliben (commented):
On my machine, `python -m unittest discover` in the root dir does not run any tests. Am I doing something wrong?

jdufresne:
Did you test using this branch or the `master` branch?

If you tested using `master`, you much first create the empty file `tests/__init__.py` so the `tests/` directory is a package. That has already been included with this PR.

eliben:
> Did you test using this branch or the `master` branch?
> 
> If you tested using `master`, you much first create the empty file `tests/__init__.py` so the `tests/` directory is a package. That has already been included with this PR.

Ah, my bad. I missed that.

eliben (approved):
LGTM, please resolve conflict due to removal of `tox.ini` and I'll merge this.

jdufresne:
done :slightly_smiling_face:

eliben (approved):

PR #401, opened 2020-12-07
Extra line breaks when c_ast.If are chained

pinzaghi:
```
>>> from pycparser import c_ast, c_generator
>>> generator = c_generator.CGenerator()
>>> generator.visit(c_ast.If(None, None, c_ast.If(None, None, c_ast.If(None, None, None))))
'if ()\n  \nelse\n  if ()\n  \nelse\n  if ()\n  \n\n\n'
```

Multiple trailing line breaks appear at the end of the code (one per if statement) when no one should be there.

### DIFF
--- a/tests/test_c_generator.py
+++ b/tests/test_c_generator.py
@@ -423,5 +423,21 @@ class TestCasttoC(unittest.TestCase):
         self.assertEqual(generator.visit(c_ast.Cast(void_type, test_fun)),
                          '(void) test_fun()')
 
+    def test_nested_else_if_line_breaks(self):
+        generator = c_generator.CGenerator()
+        test_ast1 = c_ast.If(None, None, None)
+        test_ast2 = c_ast.If(None, None, c_ast.If(None, None, None))
+        test_ast3 = c_ast.If(None, None, c_ast.If(None, None, c_ast.If(None, None, None)))
+        test_ast4 = c_ast.If(None, c_ast.Compound([]), c_ast.If(None, c_ast.Compound([]), c_ast.If(None, c_ast.Compound([]), None)))
+
+        self.assertEqual(generator.visit(test_ast1),
+                         'if ()\n  \n')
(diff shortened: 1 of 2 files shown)


eliben (commented):
Thanks for the contribution.

Could you please add a test?

pinzaghi:
Of course, you are welcome.

tests/test_c_generator.py
```diff
@@ -423,5 +423,21 @@ def test_to_type_with_cpp(self):
         self.assertEqual(generator.visit(c_ast.Cast(void_type, test_fun)),
                          '(void) test_fun()')
 
+    def test_nested_else_if_line_breaks(self):
+        generator = c_generator.CGenerator()
+        test_ast1 = c_ast.If(None, None, None)
+        test_ast2 = c_ast.If(None, None, c_ast.If(None, None, None))
+        test_ast3 = c_ast.If(None, None, c_ast.If(None, None, c_ast.If(None, None, None)))
+        test_ast4 = c_ast.If(None, c_ast.Compound([]), c_ast.If(None, c_ast.Compound([]), c_ast.If(None, c_ast.Compound([]), None)))
```
eliben, line 431: Hmm, this suggests another problem - empty bodies should result in `{}`, otherwise the code is not valid. Maybe it's for another PR, though.

eliben (approved):

tests/test_c_generator.py
pinzaghi: I didn't realize about that, I thought it was a code style decision. I will see if I can fix it in another PR.

eliben: I don't think the parser would ever produce `None` there for the `If` nodes, so this is limited to generator-only use, which is a fairly small problem (I can't imagine anyone using the code generator without using the parser). Regardless, I think the code

`if ()`

on its own is invalid in C. Even worse, it the next statement is something like:

```
if ()

unrelatedCode()
```

It will be parsed incorrectly.

PR #403, opened 2021-01-28
Moved the metadata into setup.cfg

KOLANICH:
Added pyproject.toml.
Version is fetched from git tags.


eliben (commented):
Can you provide a motivation for this change?

KOLANICH:
This way is better:

* can be parsed without code execution
* easier to read

No other benefits for this project (since `setup.py` is customized).

eliben:
Thanks for clarifying. Respectfully, I'll close this PR. I don't see sufficient value to make this change.

PR #404, opened 2021-01-28
Added .editorconfig

KOLANICH:


eliben (commented):
Sorry, I don't see why this is needed and the PR provides no motivation.

KOLANICH:
https://editorconfig.org/


>why this is needed

Some people have editor settings set to tabs. This project uses spaces. This file overrides the setting (and sets other ones too) for this project only.

PR #409, opened 2021-03-12
fixing limits in fake includes

bringlein:
The defines of limits.h in `utils/fake_libc_include/_fake_defines.h` were partly wrong or missing.


eliben (changes requested):
Thanks.

Could you order them consistently? E.g. min first, max second in each pair

bringlein:
sure, done.

eliben (approved):
Thanks!

Issue #63, opened 2014-12-18
C11 support?


refi64:
Are there any plans for pycparser to support C11, most particularly _Generic, _Static_assert, and _Noreturn?

eliben:
I will close this now since many C11 features have been merged over the years. It will be more useful to discuss specific lacking features in separate issues.

PR #428, opened 2021-08-26
Introduce partial C11 support

vit9696:
This changeset adds basic support for `_Noreturn`, `_Thread_local`, and `_Static_assert`. It additionally updates the fake libc headers to support their macro aliases (`noreturn`, `thread_local`, and `static_assert`) and fixes preprocessor tests on macOS.

references #63

### DIFF
--- a/pycparser/c_ast.py
+++ b/pycparser/c_ast.py
@@ -184,6 +184,17 @@ class ArrayDecl(Node):
         if self.dim is not None:
             yield self.dim
 
+    def __eq__(self, o):
+        if not o: return False
+        if not isinstance(o, ArrayDecl): return False
+        if self.type != o.type: return False
+        if self.dim != o.dim: return False
+        if self.dim_quals != o.dim_quals: return False
+        return True
+
+    def __ne__(self, o):
+        return not self == o
+
     attr_names = ('dim_quals', )
 
 class ArrayRef(Node):
@@ -204,6 +215,16 @@ class ArrayRef(Node):
             yield self.name
         if self.subscript is not None:
             yield self.subscript
+
+    def __eq__(self, o):
+        if not o: return False
+        if not isinstance(o, ArrayRef): return False
+        if self.name != o.name: return False
+        if self.subscript != o.subscript: return False
+        return True
+
+    def __ne__(self, o):
+        return not self == o
 
     attr_names = ()
 
@@ -227,7 +248,70 @@ class Assignment(Node):
         if self.rvalue is not None:
             yield self.rvalue
 
+    def __eq__(self, o):
+        if not o: return False
+        if not isinstance(o, Assignment): return False
+        if self.op != o.op: return False
+        if self.lvalue != o.lvalue: return False
+        if self.rvalue != o.rvalue: return False
+        return True
+
+    def __ne__(self, o):
+        return not self == o
+
     attr_names = ('op', )
+
+class Alignas(Node):
+    __slots__ = ('alignment', 'coord', '__weakref__')
+    def __init__(self, alignment, coord=None):
+        self.alignment = alignment
+        self.coord = coord
+
+    def children(self):
+        nodelist = []
+        if self.alignment is not None: nodelist.append(("alignment", self.alignment))
+        return tuple(nodelist)
+
+    def __iter__(self):
+        if self.alignment is not None:
+            yield self.alignment
+
+    def __eq__(self, o):
+        if not o: return False
+        if not isinstance(o, Alignas): return False
+        if self.alignment != o.alignment: return False
+        return True
+
+    def __ne__(self, o):
+        return not self == o
+
+    attr_names = ()
+
+class Atomic(Node):
+    __slots__ = ('subtype', 'coord', '__weakref__')
+    def __init__(self, subtype, coord=None):
+        self.subtype = subtype
+        self.coord = coord
+
+    def children(self):
+        nodelist = []
+        if self.subtype is not None: nodelist.append(("subtype", self.subtype))
+        return tuple(nodelist)
+
+    def __iter__(self):
+        if self.subtype is not None:
+            yield self.subtype
+
+    def __eq__(self, o):
+        if not o: return False
+        if not isinstance(o, Atomic): return False
+        if self.subtype != o.subtype: return False
+        return True
+
+    def __ne__(self, o):
+        return not self == o
+
+    attr_names = ()
 
 class BinaryOp(Node):
     __slots__ = ('op', 'left', 'right', 'coord', '__weakref__')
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@ parser.out
 utils/z.c
 *.egg-info
 *.swp
+.DS_Store
```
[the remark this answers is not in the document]
vit9696: This is a file generated automatically by macOS (https://en.wikipedia.org/wiki/.DS_Store).

pycparser/c_parser.py
```diff
@@ -561,6 +561,16 @@ def p_external_declaration_4(self, p):
...
 
+    def p_external_declaration_5(self, p):
+        """ external_declaration    : static_assert
+        """
+        p[0] = p[1]
+
+    def p_static_assert_declaration(self, p):
+        """ static_assert           : _STATIC_ASSERT LPAREN constant_expression COMMA unified_string_literal RPAREN
+        """
+        p[0] = p[6]
```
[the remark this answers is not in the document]
vit9696: > It's odd this assigns p[6] -- isn't that the RPAREN?

This is indeed RPAREN, but it was not odd as I did not need a dedicated node for `_Static_assert`, only to get the files parsed.

> Is there a place you test what node is created for static asserts? Should there be a new node type?

I agree that this is not notice, and updated the code to produce `StaticAssert` nodes here. I also added a test that ensures that it is the exact node created.

vit9696:
> Thanks for the contribution. Leaving some comments to consider

Thanks for the review. I believe I addressed them all.

pycparser/c_ast.py
```diff
@@ -510,6 +510,22 @@ def __iter__(self):
 
     attr_names = ('name', )
 
+class Atomic(Node):
```
eliben, line 513: There are two issues here:

1. You should not modify `c_ast.py` directly - it's an auto-generated file. Modify `_c_ast.cfg` instead and rerun `_ast_gen.py`. The README mentions this briefly
2. It's not clear to me why Atomic is a Node at all, if it's only a type qualifier or specifier (like `const` or `static`). Can't it be handled the same way?

vit9696: 1. Will check.
2. It is both a specifier and a qualifier, this node is to handle a specifier.

vit9696: Forgive my ignorance, but how can I run `c_ast.py`?

eliben: Specifiers are handled without nodes. See how `register` is handled, for example. BTW, `_Thread_local` is also a specifier (per https://en.cppreference.com/w/c/language/storage_duration) so you would have to follow the same path there.

vit9696: Sorry, but how can `_Atomic(_Atomic(int) *)` be reconstructed without a node?

eliben: > Forgive my ignorance, but how can I run `_ast_gen.py`?

My bad, the instructions are not clear about this. I've updated the `README` to be more clear. Please LMK if it works for you.

vit9696: Thank you, the instructions did indeed work for me, and I made a commit with the changes. As for `_Atomic`, I do not believe there is a way to handle them without a Node. Let me know if you think I am wrong.

eliben: I'm not deeply familiar with C11 or `_Atomic`, but it's not clear to me how this is different from `const int * const p`, where both the integer and the pointer are const. Isn't your example similar (both the integer and the pointer are atomic)?

Is the issue here because `_Atomic` could be written both ways? From reading the docs it seems they both compile to the same thing, and there's no semantic difference (except that as a specifier there are more restrictions - but pycparser typically assumes it's compiling correct code anyway).

vit9696: `_Atomic(x)` and `_Atomic x` are different things and the former is not equivalent to the latter. Several C11 specifiers, like `_Alignas(x)` or `_Atomic(x)` have arguments, and thus require nodes.

eliben: Reading http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1485.pdf, it says:

> 
> "The _Atomic keyword also can also be used in the form _Atomic(T), where T is a type, as a type specifier equiva-
> lent to _Atomic T."
> 

So semantically, they seem equivalent. The right way to have specifiers in pycparser is as the `quals` attribute on declarations, and having `Atomic` as a node will add a second way to represent something equivalent.

One issue I can see is that having it as a node makes *parsing* nested qualifiers like the one you've shown easier. This is something to think about. What I mean is that, if you have:

`_Atomic(<some decl>)`, you can return the node for `<some decl>` with the atomic qualifier added, and so on recursively. Would that work, without a special node?

FWIW I think `_Alignas` is different, because its argument has semantic meaning beyond wrapping another type.

vit9696: >> "The _Atomic keyword also can also be used in the form _Atomic(T), where T is a type, as a type specifier equivalent to _Atomic T."
>
> So semantically, they seem equivalent.

I think there are two sides here:

- It works only in one direction. I.e. every pair of qualifier-type `_Atomic T` can be written as specifier-type `_Atomic(T)`. It is not possible to do it vice-versa, i.e. not every pair of specifier-type  `_Atomic(T)` can be written as qualifier-type `_Atomic T`.
- For the AST I believe it may be important to distinguish the pair type. For instance, one can make style suggestions in this case like "Rewrite it this way or that way".

> `_Atomic(<some decl>)`, you can return the node for `<some decl>` with the atomic qualifier added, and so on recursively. Would that work, without a special node?

I do not believe it can work, because of the following example:

```C
_Atomic(int *) a;
_Atomic(int) *b;
_Atomic int *c;
```

`a` is an atomic storing `int` pointers, while `b` and `c` are pointers to atomic `int`. In my opinion, for this very reason I cannot put `_Atomic` to something like `quals` attribute in case it is a specifier, as it clearly defines a type hierarchy. I believe the closest what you have in the code is the `enum` specifier.

> FWIW I think `_Alignas` is different, because its argument has semantic meaning beyond wrapping another type.

I am not sure. Currently I am implementing `_Alignas` and have difficulties understanding how the grammar is applied. I pushed a change into https://github.com/vit9696/pycparser/tree/alignas, might you give me a hint what I am missing? It suggests that I have two types (`Invalid multiple types specified`), but I do not understand why this happens.

vit9696:
Disregard my claim regarding `_Alignas` and `_Alignof`. Both are implemented know.

eliben:
There's a bit too much here for a single PR. I'm still unsure about how `_Atomic` is represented here. Let's break this to smaller PRs, please.

Start with a PR that adds everything except `_Atomic` and `_Alignas`, and then these can be discussed separately.

vit9696:
Please check https://github.com/eliben/pycparser/pull/429. I can update this one with just the Alignas/Atomic once it gets merged.

eliben:
Now that the other PR has landed, I recommend opening an issue to discuss `_Atomic` before proceeding with a separate PR for that.

eliben:
I'm closing this PR. Please open a separate PR for `_Atomic` and yet another separate PR for the align stuff, when you're ready.

PR #435, opened 2021-09-20
Implement _Alignas and _Alignof support with tests

vit9696:
Rebased the original PR on master.

### DIFF
--- a/pycparser/_ast_gen.py
+++ b/pycparser/_ast_gen.py
@@ -86,7 +86,7 @@ class NodeCfg(object):
         src = self._gen_init()
         src += '\n' + self._gen_children()
         src += '\n' + self._gen_iter()
-
+        src += '\n' + self._gen_eq()
         src += '\n' + self._gen_attr_names()
         return src
 
@@ -131,6 +131,22 @@ class NodeCfg(object):
             src += '        return tuple(nodelist)\n'
         else:
             src += '        return ()\n'
+
+        return src
+
+    def _gen_eq(self):
+        src  = '    def __eq__(self, o):\n'
+        src += '        if not o: return False\n'
+        src += '        if not isinstance(o, {}): return False\n'.format(self.name)
+
+        if self.all_entries:
+            for entry in self.all_entries:
+                src += '        if self.{0} != o.{0}: return False\n'.format(entry)
+
+        src += '        return True\n\n'
+
+        src += '    def __ne__(self, o):\n'
+        src += '        return not self == o\n'
 
         return src
 
--- a/pycparser/_ast_gen.py
+++ b/pycparser/_ast_gen.py
@@ -86,7 +86,6 @@ class NodeCfg(object):
         src = self._gen_init()
         src += '\n' + self._gen_children()
         src += '\n' + self._gen_iter()
-        src += '\n' + self._gen_eq()
         src += '\n' + self._gen_attr_names()
         return src
 
@@ -131,22 +130,6 @@ class NodeCfg(object):
             src += '        return tuple(nodelist)\n'
         else:
             src += '        return ()\n'
-
-        return src
-
-    def _gen_eq(self):
-        src  = '    def __eq__(self, o):\n'
-        src += '        if not o: return False\n'
-        src += '        if not isinstance(o, {}): return False\n'.format(self.name)
-
-        if self.all_entries:
-            for entry in self.all_entries:
-                src += '        if self.{0} != o.{0}: return False\n'.format(entry)
-
-        src += '        return True\n\n'
-
-        src += '    def __ne__(self, o):\n'
-        src += '        return not self == o\n'
 
         return src
 
--- a/pycparser/_c_ast.cfg
+++ b/pycparser/_c_ast.cfg
@@ -179,7 +179,7 @@ TypeDecl: [declname, quals, align, type*]
 # A typedef declaration.
 # Very similar to Decl, but without some attributes
 #
-Typedef: [name, quals, align, storage, type*]
+Typedef: [name, quals, storage, type*]
 
 Typename: [name, quals, align, type*]
 
--- a/pycparser/_c_ast.cfg
+++ b/pycparser/_c_ast.cfg
@@ -25,6 +25,8 @@ ArrayRef: [name*, subscript*]
 #
 Assignment: [op, lvalue*, rvalue*]
 
+Alignas: [alignment*]
+
 BinaryOp: [op, left*, right*]
 
 Break: []
@@ -59,7 +61,7 @@ Continue: []
 # init: initialization value, or None
 # bitsize: bit field size, or None
 #
-Decl: [name, quals, storage, funcspec, type*, init*, bitsize*]
+Decl: [name, quals, align, storage, funcspec, type*, init*, bitsize*]
 
 DeclList: [decls**]
 
@@ -172,14 +174,14 @@ TernaryOp: [cond*, iftrue*, iffalse*]
 
 # A base type declaration
 #
-TypeDecl: [declname, quals, type*]
+TypeDecl: [declname, quals, align, type*]
 
 # A typedef declaration.
 # Very similar to Decl, but without some attributes
 #
-Typedef: [name, quals, storage, type*]
+Typedef: [name, quals, align, storage, type*]
 
-Typename: [name, quals, type*]
+Typename: [name, quals, align, type*]
 
 UnaryOp: [op, expr*]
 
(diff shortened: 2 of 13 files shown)


pycparser/c_lexer.py
```diff
@@ -111,7 +111,8 @@ def _make_tok_location(self, token):
 
     keywords_new = (
         '_BOOL', '_COMPLEX',
-        '_NORETURN', '_THREAD_LOCAL', '_STATIC_ASSERT', '_ATOMIC',
+        '_NORETURN', '_THREAD_LOCAL', '_STATIC_ASSERT',
+        '_ATOMIC', '_ALIGNOF', '_ALIGNAS',
```
eliben, line 115: I'm not sure I see parser tests for `_Alignof` here. Did you have any?

pycparser/_c_ast.cfg
```diff
@@ -172,14 +174,14 @@ TernaryOp: [cond*, iftrue*, iffalse*]
...
 # A base type declaration
 #
-TypeDecl: [declname, quals, type*]
+TypeDecl: [declname, quals, align, type*]
 
 # A typedef declaration.
 # Very similar to Decl, but without some attributes
 #
-Typedef: [name, quals, storage, type*]
+Typedef: [name, quals, align, storage, type*]
```
eliben, line 182: Can typedefs have _Alignas?

pycparser/_ast_gen.py
```diff
@@ -134,6 +134,22 @@ def _gen_children(self):
 
         return src
 
+    def _gen_eq(self):
```
eliben, line 137: Is this related to the PR?

pycparser/_ast_gen.py
```diff
@@ -134,6 +134,22 @@ def _gen_children(self):
...
+        src  = '    def __eq__(self, o):\n'
+        src += '        if not o: return False\n'
```
eliben, line 139: Is this comparing children?

pycparser/_c_ast.cfg
vit9696: No.

pycparser/_ast_gen.py
vit9696: Yes, required by tests.

pycparser/_ast_gen.py
vit9696: Which particular case do you mean?

vit9696: `if self.{0} != o.{0}` will trigger recursive `__ne__` → `__eq__` calls. If you mean inner members.

pycparser/_ast_gen.py
eliben: Could you please clarify which tests need this? pycparser has hundreds of AST tests which don't need this comparison, so I'm wondering what's new here. In any case I would prefer to decouple this large change from the `_Align*` PR.

tests/test_c_parser.py
```diff
@@ -618,6 +623,12 @@ def test_sizeof(self):
                         ['TypeDecl',
                             ['IdentifierType', ['int']]]]]])
 
+    def test_alignof(self):
```
eliben, line 626: I'd like to see more parser tests here.

At least one test separately for `_Alignof` without `_Alignas` (look at the existing `sizeof` tests)
A test for `_Alignas` with a true constant like `4`
A test for `_Alignas` with a type inside e.g. `_Alignas(int)`.

pycparser/c_parser.py
```diff
@@ -310,6 +310,11 @@ def _fix_decl_name_type(self, decl, typename):
         decl.name = type.declname
         type.quals = decl.quals[:]
 
+        for tn in typename:
+            if isinstance(tn, c_ast.Alignas):
+                type.align = tn
```
eliben, line 315: I'm a bit confused here: isn't align an attribute, not a child node? Should it be a child node instead if you're assigning nodes to it?

pycparser/c_parser.py
```diff
@@ -310,6 +310,11 @@ def _fix_decl_name_type(self, decl, typename):
...
+                typename.remove(tn)
```
eliben, line 316: Don't modify a list you're iterating.

eliben: Also, if we allow a list of alignment specifiers, does it show here or do we only overwrite them?

pycparser/c_parser.py
```diff
@@ -920,9 +940,20 @@ def p_specifier_qualifier_list_3(self, p):
     def p_specifier_qualifier_list_4(self, p):
         """ specifier_qualifier_list  : type_qualifier_list type_specifier
         """
-        spec = dict(qual=p[1], storage=[], type=[], function=[])
+        spec = dict(qual=p[1], alignment=[], storage=[], type=[], function=[])
```
eliben, line 943: So you really need _add_declaration_specifier here? Can't you just copy the `type`?

pycparser/c_parser.py
```diff
@@ -920,9 +940,20 @@ def p_specifier_qualifier_list_3(self, p):
...
         p[0] = self._add_declaration_specifier(spec, p[2], 'type', append=True)
 
+    def p_specifier_qualifier_list_5(self, p):
+        """ specifier_qualifier_list  : alignment_specifier
+        """
+        spec = dict(qual=[], alignment=[], storage=[], type=[], function=[])
+        p[0] = self._add_declaration_specifier(spec, p[1], 'alignment')
```
eliben, line 950: Same question here, it's not clear why you need to create an empty `spec` and pass it to _add_declaration_specifier

eliben (changes requested):

pycparser/_ast_gen.py
vit9696: This one. I am fine to create a separate PR for that.

```
======================================================================
FAIL: test_alignment (tests.test_c_generator.TestCtoC)
----------------------------------------------------------------------
  File "/Users/user/AppleDeveloper/pycparser/tests/test_c_generator.py", line 100, in test_alignment
    self._assert_ctoc_correct('_Alignas(32) int b;')
  File "/Users/user/AppleDeveloper/pycparser/tests/test_c_generator.py", line 79, in _assert_ctoc_correct
    "{!r} != {!r}".format(src, src2))
AssertionError: '_Alignas(32) int b;' != '_Alignas(32) int b;\n'
```

tests/test_c_parser.py
vit9696: Done, thanks.

pycparser/c_parser.py
vit9696: It is an attribute indeed, not a child node. This code was merely a fixup for the parser limitation. Previously one of the tests triggered an issue without this code, however, currently none does. I am in a good faith to believe that the issue was resolved more properly in one of the future iterations of the code and I did not notice it. Will simply remove.

pycparser/c_parser.py
vit9696: Fixed.

pycparser/c_parser.py
vit9696: Rebased on top of master fixing. Indeed no need.

pycparser/c_parser.py
vit9696: Sounds like literal copy-pasting from the existing code, nice catch. Also fixed.

pycparser/_ast_gen.py
eliben: Alright, let's move the equality stuff out to a separate PR that can be discussed after this one lands.

eliben (changes requested):
Thanks, this looks close just. Just remove the equality generation for now, and I'll take a final look

vit9696:
Sure, shall I comment the test part for now?

eliben:
> Sure, shall I comment the test part for now?

No, please remove everything related to generating equality and the test that requires it from the PR. I'm still not convinced this is needed at all, and wanted to separate it to its own PR so as not to block the align feature

vit9696:
> No, please remove everything related to generating equality and the test that requires it from the PR. I'm still not convinced this is needed at all, and wanted to separate it to its own PR so as not to block the align feature

Hmm, this is the whole `def test_alignment` function though. There is no single test step that passes without the `__eq__` functions. Updated the PR.

tests/test_c_generator.py
```diff
@@ -96,6 +96,27 @@ def test_complex_decls(self):
         self._assert_ctoc_correct('int test(const char* const* arg);')
         self._assert_ctoc_correct('int test(const char** const arg);')
 
+    # FIXME: These require custom equality comparison.
```
vit9696, line 99: I can remove that block if you think it is better.

eliben (approved):

eliben:
I've re-enabled the commented-out tests in d4c2922407c486da1ba9c6cd769990cc64308be6

eliben:
Thank you for pushing the align and other changes through and improving C11 support for pycparser :)

PR #436, opened 2021-09-23
Add support for Python 3.10

hugovk:
Due out in a couple of weeks, on 2021-10-04:

https://www.python.org/dev/peps/pep-0619/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9", "3.10-dev"]
+        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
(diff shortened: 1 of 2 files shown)


hugovk:
By the way, there's an estimated $251.43/month available for the maintainers of pycparser:

* https://tidelift.com/lifter/search/pypi/pycparser

eliben:
I'm OK with waiting for 3.10 to be properly released, at which point presumably we don't need the `-dev` postfix?

hugovk:
Yep, after the full release we can drop the `-dev`.

Fine by me to wait, although the benefit of testing now on `3.10-dev` is it prevents merging incompatible PRs now. Helps avoid surprises later.

(For packages that provide compiled binary wheels, such as NumPy, Pandas and Pillow, there's a strong argument for releasing for 3.10 before the official release date, because it helps dependant packages test much more easily. And they have. But for pure Python packages, it's not so important.)

eliben:
Seems like Python 3.10 was released. If Github Actions supports 3.10 in its workflows, please update the PR and ping it - I will merge

hugovk:
GHA now supports 3.10 final, updated - ping @eliben!

PR #441, opened 2021-10-12
Fix examples/func_defs_add_param.py: add missing `align` parameter

e-kwsm:
https://github.com/eliben/pycparser/blob/d554122e2a5702daeb68a3714826c1c7df8cbea3/pycparser/c_ast.py#L992-L994

I’m not sure if `None` is appropriate, though.


e-kwsm:
> For consistency I think it'd be better for this to be `[]`

Fixed.

PR #442, opened 2021-10-12
Fix Flake8 Errors

e-kwsm:
- Fix PEP8 [E101](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)
- Fix PEP8 [E711](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)
- Fix E999 (python2)
- Fix PEP8 [F401](https://flake8.pycqa.org/en/4.0.1/user/error-codes.html)

### DIFF
--- a/examples/dump_ast.py
+++ b/examples/dump_ast.py
@@ -14,7 +14,7 @@ import sys
 # your site-packages/ with setup.py
 sys.path.extend(['.', '..'])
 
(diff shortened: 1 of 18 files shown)


e-kwsm:
> Can't change inside `ply` -- please revert those

Reverted.

PR #443, opened 2021-10-14
Fix typos

kianmeng:


pycparser/ply/yacc.py
```diff
@@ -32,7 +32,7 @@
 # -----------------------------------------------------------------------------
 #
 # This implements an LR parser that is constructed from grammar rules defined
-# as Python functions. The grammer is specified by supplying the BNF inside
```
eliben, line 35: Please revert all changes in the `ply/` directory - it's a vendored dependency and I don't want any changes in there

utils/benchmark/inputs/sqlite-btree.c.ppout
```diff
@@ -3652,7 +3652,7 @@ struct Btree {
   sqlite3 *db;
   BtShared *pBt;
   u8 inTrans;
-  u8 sharable;
```
eliben, line 3655: Revert this file too - it's vendored in

PR #449, opened 2021-12-10
Add missing SCHAR limit defines 

matamegger:
When I tried to use the fake libc includes I found that the limits define for `SCHAR_MAX` and `SCHAR_MIN` are missing.

This PR adds a definition for `SCHAR_MIN` as well as `SCHAR_MAX`.


eliben (approved):
Thanks for the contribution

Kusalananda4329:
- hghghkgjjhjgjkbjk/ #477 [@[](url)](`

> _**### url**_

`)

PR #471, opened 2022-09-06
Fix C-preprocessor tests on Darwin

vit9696:
This was reverted somewhere in the process of refactoring, but should work fine now. On macOS `cpp` cannot be used with comments, so calling `gcc -E` is needed.

### DIFF
--- a/examples/func_defs.py
+++ b/examples/func_defs.py
@@ -18,6 +18,7 @@ import sys
 sys.path.extend(['.', '..'])
 
 from pycparser import c_ast, parse_file
+import platform
 
 
 # A simple visitor for FuncDef nodes that prints the names and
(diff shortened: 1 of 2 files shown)


tests/test_util.py
```diff
@@ -15,7 +15,7 @@
 
 def cpp_supported():
     """Is cpp (the C preprocessor) supported as a native command?"""
-    return platform.system() == 'Linux'
```
eliben, line 18: This was specifically removed last year due to failures reported on some Macs that don't have gcc/cpp installed by default. It's not clear to me that this should change now

examples/func_defs.py
```diff
@@ -28,10 +29,16 @@ def visit_FuncDef(self, node):
...
     # make sure one exists in PATH.
-    ast = parse_file(filename, use_cpp=True,
-                     cpp_args=r'-Iutils/fake_libc_include')
+    if platform.system() == 'Darwin':
+        ast = parse_file(filename, use_cpp=True,
+                 cpp_path='gcc',
+                 cpp_args=['-E', '-Iutils/fake_libc_include'])
+    else:
+        ast = parse_file(filename, use_cpp=True,
+                         cpp_args=r'-Iutils/fake_libc_include')
```
eliben, line 41: I'm not sure I follow, sorry. You enable cpp_supported for Darwin lower down, which causes a bunch of tests to execute cpp but here you specifically use gcc -E for Darwin?

eliben (changes requested):

tests/test_util.py
vit9696: On macOS if `cpp` is installed, then `gcc` is installed too. Python on macOS comes with Xcode tools, and is not bundled, so I do not think you can install the former without the latter.

examples/func_defs.py
vit9696: Are there `cpp` calls below on macOS? Basically `cpp` on macOS does not handle C comments, so if by chance cpp is used in other examples, they might work ok if they have no C-style comments.

eliben:
This PR mixes different things: one is test utilities, another is a specific example.

For test utilities, I'm happy to keep these tests Linux-only; the tests don't really exercise a core part of the package, only a small preprocessor integration. Dealing with different variants of what's installed on what machine is too much complexity.

As for the example change, there's already another example that uses `gcc -E` instead of `cpp`. These examples are very simple and should be changed as needed. Examples are not currently tested on non-Linux.

Issue #479, opened 2022-11-07
c-to-c breaks with multiple pragma in if/for/while without compound statement


ldore-ks:
When parsing, then generating C, the following code:

```c
int glob = 0;
void f(int test) {
  if (test)
    #pragma clang foo
    #pragma clang bar
    glob += 1; // <-- Only increment glob if test is not false
}
```

Gets transformed to:

```c
int glob = 0;
void f(int test) {
  if (test) {
    #pragma foo
    #pragma bar
  }
  glob += 1; // <-- Always increment glob
}
```

Source input and output have different runtime behavior.

PR #480, opened 2022-11-08
Fix multi-pragma/single statement blocks (#479)

ldore:


tests/test_c_generator.py
```diff
@@ -339,10 +339,14 @@ def test_pragma(self):
             void f() {
                 #pragma bar
                 i = (a, b, c);
+                if (d)
```
eliben, line 342: Can you please add separate tests here with 1 pragma in the block, and two pragmas in the block.

eliben (changes requested):
Thanks. Also modify the commit/description when you rebase to say it fixes the issue you opened - so the issue will be automatically closed

tests/test_c_generator.py
ldore: Done.
Rebase comment should also be ok.

eliben (approved):

PR #482, opened 2022-11-14
Fix #478 (pragma-before-semicolon)

### LINKED ISSUE #478: #Pragmas in statements raise error
Pycparser raises a parser error when trying to parse statements that contain `#pragma`.  
A typical example comes from Apple's `iPhoneOS.sdk/usr/include/sys/event.h:371` (SDK 13).  

The relevant `sys/event.h:371` source code is:
```c
struct knote;
SLIST_HEAD(klist, knote);
```
Once pre-processed, the `SLIST_HEAD` macro expands to:

```c
struct knote;
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wnullability-completeness"
 struct klist { struct knote *slh_first; }
#pragma clang diagnostic pop
 ;
```

Note the `#pragma` that comes _before_ the end-of-statement `;`.

When parsing, pycparser raises the following error:
```
pycparser.plyparser.ParseError: Platforms/iphoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/sys/event.h:371:2: before: pragma
```

ldore:
This makes code written as
```c
a += 1
#pragma foo
;
```
to be parsed as
```c
a += 1
;
#pragma foo
```

Handling `#pragma` (and especially corner-cases) with minimal impact is tricky.  
This focuses _only_ on the `#pragma`-before-`;` case, by generating the `#pragma` pseudo-statement _after_ the real statement.

`p_pragmacomp_or_statement` covers the most common use-cases.  
For `p_statement`'s possible sub-rules:
- `p_labeled_statement`: Already handles `#pragma` before `;` through the `pragmacomp_or_statement` rule.
- `p_expression_statement`: Added sub-rule.
- `p_compound_statement`: Not relevant.
- `p_selection_statement`: Already handles `#pragma` before `;` through the `pragmacomp_or_statement` rule.
- `p_iteration_statement`: Several sub-rules already handle `#pragma` through the `pragmacomp_or_statement` rule which is already fixed by #479.
   Tweaks added to `p_continue`, `p_break`, `p_return`.
- `p_jump_statement`: Modified in a similar way as `p_expression_statement`
- `pppragma_directive`: Not relevant
- `static_assert`: Not relevant

Notes:
- I'm not sure if the modified rules should be written as an additional line, or with an `_opt` suffix.  
  E.g. Would you recommend writing:
  ```python
    def p_jump_statement_2(self, p):
        """ jump_statement  : BREAK SEMI 
                            | BREAK pppragma_directive_list SEMI
        """
  ```
  or rather change to something like:
  ```python
    def p_jump_statement_2(self, p):
        """ jump_statement  : BREAK pppragma_directive_list_opt SEMI """
  ```
  Does one style have an impact on parsing (performance, readability...) ?

- Similarly, for rules that rely on `len(p)`, would you recommend writing rules as:
  ```python
    def p_jump_statement_5(self, p):
        """ jump_statement  : RETURN expression pppragma_directive_list SEMI
                            | RETURN pppragma_directive_list SEMI
        """
        p[0] = [c_ast.Return(p[2] if len(p) == 5 else None, self._token_coord(p, 1))] + p[3 if len(p) == 5 else 2]
  ```
  or split the rule to _not_ rely on `len(p)` as :
  ```python
    def p_jump_statement_5(self, p):
        """ jump_statement  : RETURN expression pppragma_directive_list SEMI
        """
        p[0] = [c_ast.Return(p[2], self._token_coord(p, 1))] + p[3]

    def p_jump_statement_6(self, p):
        """ jump_statement  : RETURN pppragma_directive_list SEMI
        """
        p[0] = [c_ast.Return(None, self._token_coord(p, 1))] + p[2]
  ```

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -480,6 +480,13 @@ class CParser(PLYParser):
         else:
             return c_ast.Union
 
+    def _to_list(self, p):
+        """ Normalize the input to a list: if given a list, return as-is, 
+            else return single-element list.
+        """
+        return p if isinstance(p, list) else [p]
+
+
     ##
     ## Precedence and associativity of operators
     ##
@@ -682,10 +689,10 @@ class CParser(PLYParser):
         """
         if len(p) == 3:
             p[0] = c_ast.Compound(
-                block_items=p[1]+[p[2]],
+                block_items=p[1] + self._to_list(p[2]),
                 coord=self._token_coord(p, 1))
         else:
-            p[0] = p[1]
+            p[0] = c_ast.Compound(block_items=p[1], coord=self._token_coord(p, 1)) if isinstance(p[1], list) else p[1]
 
     # In C, declarations can come several in a line:
     #   int x, *px, romulo = 5;
@@ -759,8 +766,9 @@ class CParser(PLYParser):
     # line is reached.
     def p_declaration(self, p):
         """ declaration : decl_body SEMI
+                        | decl_body pppragma_directive_list SEMI
         """
-        p[0] = p[1]
+        p[0] = p[1] if len(p) == 3 else self._to_list(p[1]) + p[2]
 
     # Since each declaration is a list of declarations, this
     # rule will combine all the declarations and return a single
@@ -1529,7 +1537,7 @@ class CParser(PLYParser):
         """ block_item  : declaration
                         | statement
         """
-        p[0] = p[1] if isinstance(p[1], list) else [p[1]]
+        p[0] = self._to_list(p[1])
 
     # Since we made block_item a list, this just combines lists
     #
@@ -1563,7 +1571,7 @@ class CParser(PLYParser):
         p[0] = c_ast.If(p[3], p[5], None, self._token_coord(p, 1))
 
     def p_selection_statement_2(self, p):
-        """ selection_statement : IF LPAREN expression RPAREN statement ELSE pragmacomp_or_statement """
+        """ selection_statement : IF LPAREN expression RPAREN pragmacomp_or_statement ELSE pragmacomp_or_statement """
         p[0] = c_ast.If(p[3], p[5], p[7], self._token_coord(p, 1))
 
     def p_selection_statement_3(self, p):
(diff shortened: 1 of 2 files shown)


pycparser/c_parser.py
```diff
@@ -480,6 +480,13 @@ def _select_struct_union_class(self, token):
         else:
             return c_ast.Union
 
+    def _to_list(self, p):
+        """ Normalize the input to a list: if given a list, return as-is,
```
eliben, line 484: I'd rather have this refactoring in a separate PR that can come earlier. It obscures the actual changes in this PR

eliben (commented):
It's not clear to me what's the canonical way to do this

Is this defined in the standard? How do other compilers handle this case?

ldore:
About
> It's not clear to me what's the canonical way to do this
> Is this defined in the standard?

No. This is a workaround.  
Proper handling of `#pragma` would preserve `#pragma` _before_ the `;` token.

> How do other compilers handle this case?

`#pragma` are preprocessor directives, and don't seem to appear in `gcc`/`clang` compiler ASTs.  
Source rewriters _may_ work on an AST (but I couldn't find one in `astyle` and `clang-format`).

I checked on a simple main function containing a few `#pragma`:
```c
int main(int argc, const char * argv[]) {
  #pragma clang diagnostics push
  #pragma clang diagnostic ignored "-Wunused-variable"
  return 0;
  #pragma clang diagnostics pop
}
```
(Negative) results:
- llvm/clang does not show `#pragma` in its AST dump (ran: `clang -cc1 -ast-dump test.c`)
- llvm/clang does not mention `#pragma` in its bitcode conversion either (ran: `clang -S -emit-llvm test.c`)  
  llvm documentation mentions implementing [`PragmaHandler`](https://clang.llvm.org/doxygen/classclang_1_1PragmaHandler.html)s but that's for the preprocessor.
- GCC does not have the `pragma` string anywhere in its dump files (ran: `gcc -c -fdump-tree-all-graph test.c`)
- antlr only creates a parse tree. It's up to the user to define the AST, and I couldn't find an example project with one.

eliben:
Thanks for trying other compilers/tools.

What worries me about this direction is that it's a hole with no bottom. Theoretically, just like comments, pragmas can appear *anywhere*. E.g. your example:

```
a += 1
#pragma foo
;
```

But this could also be:

`
a +=
#pragma foo
1;
`

And so on. So it means that we should be able to accommodate for pragmas basically at every token in the parser. With the current structure of pycparser (using a Yacc-like grammar with PLY) this is an enormous increase in complexity for an odd corner case. This is why pycparser generally only handles preprocessed code and doesn't deal with comments etc. I started supporting basic pragmas for some more common scenarios, but I'm not sure we should continue down this road.

eliben:
Closing this PR due to inactivity for now. Feel free to reopen if you have time to pick it back up

PR #486, opened 2023-01-05
Add an encoding parameter to parse_file

jordr:
We may need to process files written on/output by compilers on a system with a different [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding), which is the default of [open()](https://docs.python.org/3/library/functions.html#open).

This patch exposes the `encoding` optional parameter of `open` to the `parse_file` API.


eliben (changes requested):
Thanks. Please rebase on master to ensure tests are passing

eliben (approved):

Issue #485, opened 2023-01-02
_Pragma is not supported


jordr:
The use of `_Pragma` instead of `#pragma` is accepted in C99. [1]
pycparser raises a ParseError on this minimal main.c:
```c
/* main.c */
_Pragma ("GCC dependency \"parse.y\"")

int main() {
        return 0;
}
```

Python3+pycparser prompt:
```py
from pycparser import parse_file, c_generator
parse_file('./main.c', use_cpp=False)
```

Output:
```sh
>>> parse_file('./main.c', use_cpp=False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jordy/.local/lib/python3.8/site-packages/pycparser/__init__.py", line 90, in parse_file
    return parser.parse(text, filename)
  File "/home/jordy/.local/lib/python3.8/site-packages/pycparser/c_parser.py", line 147, in parse
    return self.cparser.parse(
  File "/home/jordy/.local/lib/python3.8/site-packages/pycparser/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/home/jordy/.local/lib/python3.8/site-packages/pycparser/ply/yacc.py", line 1199, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/home/jordy/.local/lib/python3.8/site-packages/pycparser/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/home/jordy/.local/lib/python3.8/site-packages/pycparser/c_parser.py", line 1931, in p_error
    self._parse_error(
  File "/home/jordy/.local/lib/python3.8/site-packages/pycparser/plyparser.py", line 67, in _parse_error
    raise ParseError("%s: %s" % (coord, msg))
pycparser.plyparser.ParseError: ./main.c:1:10: before: "GCC dependency \"parse.y\""
```

[1] https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html

JulianKemmerer:
Interesting - maybe a regression or something I am using pycparser right now to parse `_Pragma`s just fine
(though surely my version is quite old, ~2018?)

my code where I use it (string replacement for macro stuff that doesnt work with 
`#pragma THING` syntax doesnt work where `THING` could be another macro etc.)

```c
#define PRAGMA_MESSAGE(x) _Pragma(#x)
[...]

PR #487, opened 2023-01-07
Feature/add  pragma support

jordr:
Resolves https://github.com/eliben/pycparser/issues/485 for [_Pragma](https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html) support.
After the preprocessing stage, some `_Pragma`s may remain. This at least lets us parse such code. 

I'm not certain about the `_token_coord(p, 2))` here though. Might be 3?
https://github.com/statinf-software/pycparser/blob/32b35034bf06d76af10054da8cc227b9e6dd75dd/pycparser/c_parser.py#L579-L580

(PS: I get `ModuleNotFoundError: No module named '_ast_gen'` when trying to run the tests, but I think the test additions still ran and passed.)

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -571,6 +571,11 @@ class CParser(PLYParser):
         self._parse_error('Directives not supported yet',
                           self._token_coord(p, 1))
 
+    # This encompasses two types of C99-compatible pragmas:
+    # - The #pragma directive:
+    #       # pragma character_sequence
+    # - The _Pragma unary operator:
+    #       _Pragma ( " string_literal " )
     def p_pppragma_directive(self, p):
         """ pppragma_directive      : PPPRAGMA
                                     | PPPRAGMA PPPRAGMASTR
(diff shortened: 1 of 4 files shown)


pycparser/c_parser.py
```diff
@@ -574,8 +574,11 @@ def p_pp_directive(self, p):
     def p_pppragma_directive(self, p):
         """ pppragma_directive      : PPPRAGMA
                                     | PPPRAGMA PPPRAGMASTR
+                                    | _PRAGMA LPAREN unified_string_literal RPAREN
```
eliben, line 577: Can you add a comment explaining why the contents of `_PRAGMA` differ from `PPPRAGMA`?

eliben (changes requested):
Thank you; please rebase this PR on top of the current master - it had a commit that should fix action failures on old Python versions

pycparser/c_parser.py
jordr: Well, it has a different syntax? 
image

`#pragma` does not require parenthesis and strings as far as I know. Not sure what to explain?

eliben: The comment is good, thanks! Please move it out of the docstring though - this doesn't work with PLY's yacc (see test error in the GitHub action)

eliben (changes requested):

jordr:
Thanks for the help! Looks good to me now.

eliben (approved):

PR #492, opened 2023-02-23
Set up permissions to ci.yml

### LINKED ISSUE #491: Set up permissions to github workflows
Hi, I'm from google and the openSSF and I'm working on improving supply-chain security on many open sources.

I would like to suggest to set the GITHUB_TOKEN permissions of your workflows as read only in the [top level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) and grant any write permission needed at the [run level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions).

It is a default behavior of github workflows to grant write permissions to all permissions, thus it is both a recommendation from [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) to always use credentials that are minimally scoped.

I'll send a PR just to show what the changes are about, but fell free to reach me out in case of any doubts or concerns about it, and I hope I can help pycparser to increase its supply-chain security even more.

joycebrum:
Closes #491 

Changes:
- set top level permission to contents: read

Test:
It ran successfully https://github.com/joycebrum/pycparser/actions/runs/4255961081

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
+
 jobs:
   build:
 


eliben (approved):
Thanks!

PR #499, opened 2023-03-17
Create a Security Policy

### LINKED ISSUE #498: Create a Security Policy
A Security Policy is a GitHub standard document (`SECURITY.md`) that can be seen in the “Security Tab”  to instruct users about how to report vulnerability in the safest and most efficient way possible.

image

It is a [Scorecard Recommendation](https://github.com/ossf/scorecard/blob/main/docs/checks.md#security-policy) (being one check of medium priority) and a [Github Recommendation](https://docs.github.com/en/code-security/getting-started/securing-your-repository#setting-a-security-policy).

Together with this issue I’ll submit one suggestion of Security Policy, feel free to edit it directly or ask me for editions until it is in compliance with how pycparser would best handle with vulnerabilities reports.

joycebrum:
Closes #498 

I’ve created the SECURITY.md file considering the [report vulnerability through security advisory](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability), which is a new github feature still in beta and that has to be enabled.

If you rather not enabling it there is also the possibility to receive the vulnerability report through an email, in this case just let me know which email it would be and I’ll submit the change.

Besides that feel free to edit or suggest any changes to this document, it is supposed to reflect the amount of effort the team can offer to handle vulnerabilities.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+To report a security issue, please disclose it at [security advisory](https://github.com/eliben/pycparser/security/advisories/new).
+
+We will respond within 7 working days of your submission. If the issue is confirmed as a vulnerability, we will open a Security Advisory and acknowledge your contributions as part of it. This project follows a 90 day disclosure timeline.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,4 +2,4 @@
 
 To report a security issue, please disclose it at [security advisory](https://github.com/eliben/pycparser/security/advisories/new).
 
-We will respond within 7 working days of your submission. If the issue is confirmed as a vulnerability, we will open a Security Advisory and acknowledge your contributions as part of it. This project follows a 90 day disclosure timeline.
+We will respond within 14 working days of your submission. If the issue is confirmed as a vulnerability, we will open a Security Advisory and acknowledge your contributions as part of it. This project follows a 90 day disclosure timeline.


eliben (changes requested):
I'm not sure what the 90 day disclosure is and how it applies here. 

14 days would be better than 7

joycebrum:
The 90 days to disclosure is regarding the [Public Vulnerability Disclosure](https://www.techtarget.com/searchsecurity/definition/vulnerability-disclosure#:~:text=What%20is%20vulnerability%20disclosure%3F,in%20computer%20software%20or%20hardware.), which is basically when the vulnerability will be made public to users know about it and work in their own remediations.

In github there is the [Security Advisory](https://github.com/eliben/pycparser/security/advisories) which is used to do this public vulnerability disclosure.

AFAIK, there is some possible standard rules a project/org may adopt, which 90 days to disclosure is one of them:

- until it is fixed;
- until a certain amount of time has passed since a report was first submitted; (the 90 days for example)

According to [tech target blog](https://www.techtarget.com/searchsecurity/definition/vulnerability-disclosure#:~:text=What%20is%20vulnerability%20disclosure%3F,in%20computer%20software%20or%20hardware.), most industry vendors, as well as Google's Project Zero team, recommend a 90-day deadline to fix a vulnerability before full public disclosure, with a seven-day requirement for critical security issues but fewer than seven days for critical vulnerabilities being actively exploited.

Let me know if you want to keep the 90 days of public disclosure or if you rather remove this part of the document and perhaps use the "until it is fixed" rule.

eliben (approved):

Issue #505, opened 2023-05-17
Hash Pin Github Action on Workflows


joycebrum:
## Description

There is another security practice I would like to suggest, also recommended by the [OpenSSF Scorecard][scorecard-repo], which is to hash pin dependencies to prevent typosquatting and tag renaming attacks. 

The change would only be applied to GitHub workflows.

This means:
- Hash pinning GitHub Workflow actions.
```yml
actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
```

Along with hash-pinning dependencies, I also recommend adopting dependabot or renovatebot to help keep the dependencies up to date. Both tools can update  hashes and associated semantic version comments. 

Together with this issue I'll also suggest a PR with the hash-pinning changes since they are quite simple. If you also agree on adopting dependency update tool just tell me which one and I can submit a default config file that covers github workflow permissions.

Any questions or concerns just let me know.
Thanks!

## Additional Context

A tag renaming attack is a type of attack whereby an attacker:
- Hijack an action.
- Upload a malicious version.
- Replace existing tags with malicious versions.

 
A [typosquatting attack][typosquatting] is a type of attack whereby an attacker:
- Create a malicious package
- Publish it with a similar name of a known package (example: actiona instead of actions)

For more informations about the dependency-update tools:
- [Dependabot][dependabot]
- [Renovatebot][renovatebot]

[scorecard-repo]: https://github.com/ossf/scorecard
[dep-confusion]: https://www.websecuritylens.org/how-dependency-confusion-attack-works-and-how-to-prevent-it/ 
[typosquatting]: https://snyk.io/blog/typosquatting-attacks/ 
[dependabot]: https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/
[renovatebot]:https://www.mend.io/renovate/

eliben:
Closed per discussion in #506

PR #506, opened 2023-05-17
Hash pin ci.yml workflow

joycebrum:
Issue #505 

Example of success run with hash pin https://github.com/joycebrum/pycparser/actions/runs/5006562582

Hashes:

- [8e5e7e5ab8b370d6c329ec480221332ada57f0ab](https://github.com/actions/checkout/commit/8e5e7e5ab8b370d6c329ec480221332ada57f0ab)
- [57ded4d7d5e986d7296eab16560982c6dd7c923b](https://github.com/actions/setup-python/commit/57ded4d7d5e986d7296eab16560982c6dd7c923b)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,9 @@ jobs:
 
     steps:
 
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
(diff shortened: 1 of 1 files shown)


eliben (changes requested):
I don't understand -- is the concern here that an attacker will hijack the official `actions/checkout` action?

joycebrum:
Yeah the attacker can hijack or tag rename any version if the action repo got compromised.

Although the actions are official they are also open source projects and unfortunatelly can be compromised as any other project could.

But it is important to noticed that this is just an extra precaution (since the token permission is already read only, no much harm could be done).

eliben:
Thanks for the clarification. I think I'll prefer to keep it simple for now

PR #507, opened 2023-05-29
New example to generate AST from scratch

Andree37:
As I was doing some research and I came along this 

I found there was no example on how to create an AST from scratch and print out the target C code.

I purpose something of this sort as an initial example. As I am a beginner of the module, please don't be afraid to suggest changes :)

### DIFF
--- a/examples/generate_ast.py
+++ b/examples/generate_ast.py
@@ -1,0 +1,44 @@
+# -----------------------------------------------------------------
+# pycparser: generate_ast.py
+#
+# Tiny example of writing an AST from scratch to C code.
+#
+# Andre Ribeiro [https://github.com/Andree37]
+# License: BSD
+# -----------------------------------------------------------------
+from __future__ import print_function
+
+from pycparser import c_ast, c_generator
+
+
+# target C code:
+# int main() {
+#     return 0;
+# }
+
+
+def create_your_ast():
+    constant_zero = c_ast.Constant(type='int', value='0')
+    return_node = c_ast.Return(expr=constant_zero)
+    compound_node = c_ast.Compound(block_items=[return_node])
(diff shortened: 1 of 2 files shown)


examples/generate_ast.py
```diff
@@ -0,0 +1,44 @@
+# -----------------------------------------------------------------
+# pycparser: generate_ast.py
+#
+# Tiny example of writing an AST from scratch to C code.
+#
+# Andre Ribeiro [https://github.com/Andree37]
+# License: BSD
+# -----------------------------------------------------------------
+from __future__ import print_function
```
eliben, line 9: This is unnecessary now -- I also just removed it from other examples for consistency

examples/generate_ast.py
```diff
@@ -0,0 +1,44 @@
...
+# int main() {
+#     return 0;
+# }
+
+
+def create_your_ast():
+    constant_zero = c_ast.Constant(type='int', value='0')
+    return_node = c_ast.Return(expr=constant_zero)
+    compound_node = c_ast.Compound(block_items=[return_node])
+    type_decl_node = c_ast.TypeDecl(declname='main', quals=[], type=c_ast.IdentifierType(names=['int']), align=[])
```
eliben, line 24: pycparser code typically tries to wrap at 80 columns
should be easy here with all the keyword params (like you do in func_def_node)

eliben (changes requested):
Thanks. I suggest a different file name because "generate" is ambiguous here (this library is all about generating ASTs from C source): `examples/construct_ast_from_scratch.py`

Andree37:
Corrected :)

eliben (approved):

eliben:
Some of the builders are failing - check out other examples to see how to set up the import path to work properly.

Andree37:
> Some of the builders are failing - check out other examples to see how to set up the import path to work properly.

I believe I have corrected it now, unless I also needed to `import platform`

PR #518, opened 2023-09-04
Beautify Node.show() for parse tree printing 

Dob-The-Duilder:
Rather than just indents, use an actual tree structure for better displaying syntax trees.

Before
image

After:
image

### DIFF
--- a/pycparser/_ast_gen.py
+++ b/pycparser/_ast_gen.py
@@ -224,12 +224,15 @@ class Node(object):
         """
         pass
 
-    def show(self, buf=sys.stdout, attrnames=False, nodenames=False, showcoord=False, _my_node_name=None, indent='', islast = True):
+    def show(self, buf=sys.stdout, offset=0, attrnames=False, nodenames=False, showcoord=False, _my_node_name=None, indent='', islast = True):
         """ Pretty print the Node and all its attributes and
             children (recursively) to a buffer.
 
             buf:
                 Open IO buffer into which the Node is printed.
+
+            offset:
+                Initial offset (amount of leading spaces)
 
             attrnames:
                 True if you want to see the attribute names in
@@ -250,7 +253,7 @@ class Node(object):
                 Denotes if node is final element in its tree.
         """
         marker = "└─" if islast else "├─"
-        lead = indent + marker
+        lead = ' ' * offset + indent + marker
         
         if nodenames and _my_node_name is not None:
             buf.write(lead + self.__class__.__name__+ ' <' + _my_node_name + '>: ')
@@ -275,6 +278,7 @@ class Node(object):
         for idx, (child_name, child) in enumerate(self.children()):
             child.show(
                 buf,
+                offset=offset,
                 attrnames=attrnames,
                 nodenames=nodenames,
                 showcoord=showcoord,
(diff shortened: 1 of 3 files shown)


eliben (changes requested):
This file is auto-generated -- see comment at the top. This change should be made in the generator -- `_ast_gen.py`

Dob-The-Duilder:
My mistake. Change has now been made in the _ast_gen.py file.

eliben:
Tests are failing

pycparser/c_ast.py
```diff
@@ -76,8 +73,16 @@ def show(self, buf=sys.stdout, offset=0, attrnames=False, nodenames=False, showc
             showcoord:
                 Do you want the coordinates of each Node to be
                 displayed.
+
+            indent:
+                Spaces and lines to be displayed behind node.
```
eliben, line 78: I don't understand what this means -- "displayed behind node"?

pycparser/c_ast.py
```diff
@@ -96,14 +101,23 @@ def show(self, buf=sys.stdout, offset=0, attrnames=False, nodenames=False, showc
             buf.write(' (at %s)' % self.coord)
         buf.write('\n')
 
+        indent += "  " if islast else "│ "
+        
+        lastChild = None
+        
+        children = self.children()
+        if len(children) > 0:
+            _, lastChild = children[-1]
```
eliben, line 110: I think you should be able to accomplish this in a cleaner way using `enumerate` in the `for` loop below. Then you can trivially find if the index is the one before last and you won't need this condition and `lastChild`

eliben (changes requested):

pycparser/c_ast.py
Dob-The-Duilder: Left over from older version, forgot to change. As for the failing tests I accidentally typed an 's' near the top of the file

pycparser/c_ast.py
Dob-The-Duilder: I think I understood what you meant but if not let me know

pycparser/_ast_gen.py
```diff
@@ -245,8 +242,16 @@ def show(self, buf=sys.stdout, offset=0, attrnames=False, nodenames=False, showc
             showcoord:
                 Do you want the coordinates of each Node to be
                 displayed.
+
+            indent:
+                Spaces and verticle lines to be displayed infront of node.
```
eliben, line 247: vertical
in front

pycparser/_build_tables.py
```diff
@@ -20,7 +20,7 @@
 # Generate c_ast.py
 from _ast_gen import ASTCodeGenerator
 ast_gen = ASTCodeGenerator('_c_ast.cfg')
-ast_gen.generate(open('c_ast.py', 'w'))
+ast_gen.generate(open('c_ast.py', 'w', encoding='UTF-8')))
```
eliben, line 23: Why is this change in this PR?

eliben (changes requested):

eliben:
Added some more minor comments; also - all the tests fail on CI, please fix

pycparser/_build_tables.py
Dob-The-Duilder: it raises the following error when building without that part:
`UnicodeEncodeError: 'charmap' codec can't encode characters in position 2511-2512: character maps to <undefined>`
Those characters end up corresponding to '└' and '─'. No idea why as it should default to UTF-8. but adding the encoding specifier fixes it.

PR #535, opened 2024-04-23
Comments, directive, unknown function, and variable work around

aerah8:
The main purpose of the preprocessing file is to get rid of all comments and work around directive files. This increases the versatility of the parser function.


eliben:
Thanks, but I'm not going to add preprocessing to pycparser at this time. It's a much more difficult problem than a simple patch could handle. pycparser requires preprocessed code: https://github.com/eliben/pycparser?tab=readme-ov-file#interaction-with-the-c-preprocessor

Issue #539, opened 2024-05-01
Can't parse function declaration with unnamed static array


gperciva:
Thanks for this library!  I'm having great fun using it to nitpick some code.

Attempting to parse a function declaration with an unnamed static array:
```
from pycparser.c_parser import CParser
CParser().parse('void f(int [static 10]);')
```
fails with:

```
  File "/usr/local/lib/python3.9/site-packages/pycparser/plyparser.py", line 67, in _parse_error
    raise ParseError("%s: %s" % (coord, msg))
pycparser.plyparser.ParseError: :1:13: before: static
```

If we name the argument -- i.e. `(int x[static 10])` -- then it works fine (since the example would be exactly the same as #21, the issue which prompted support for static arrays in the first place).  Omitting the argument name is valid in C99.

I'm using pycparser 2.22.  Thanks again for this library!

eliben:
PRs to add support are welcome :)

Issue #21, opened 2014-01-16
Can't parse array parameter with 'static'


rmartinjak:
The following code results in a ParseError, even though it is valid C99, according to http://www.iso-9899.info/n1256.html#6.7.5.3p7

```
from pycparser.c_parser import CParser
CParser().parse('void f(int x[static 10]);')
```

PR #545, opened 2024-06-19
Allow abstract declarator "static" (GH issue #539)

gperciva:
This is similar to:

>    allow "static" in array parameters (GH issue #21)
>    aac7b27d7378cf45424b1f5f1b2b450c62636dde

which was revised shortly after in:

>    Fuller support for qualifiers in array dimensions.
>    8aad3186f39127ec9544f2b8c412de2bb7300fd4

The grammar is as defined in C99 6.7.6 Type names, or A.2.2 Declarations (6.7.6).

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -1530,6 +1530,20 @@ class CParser(PLYParser):
             type=c_ast.TypeDecl(None, None, None, None),
             coord=self._token_coord(p, 1))
 
+    def p_direct_abstract_declarator_8(self, p):
+        """ direct_abstract_declarator  : LBRACKET STATIC type_qualifier_list_opt assignment_expression RBRACKET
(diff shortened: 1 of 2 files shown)


gperciva:
I debated between adding `p_direct_abstract_declarator_8()`, vs. stuffing it in `p_direct_abstract_declarator_3()`.

Putting it in `_3` would keep the grammar in the same order as 6.7.6 -- namely, the `static` functions going between `direct-abstract-declarator_opt [ type-qualiﬁer-list_opt
assignment-expression_opt ]` and `direct-abstract-declarator_opt [ * ]`.  However, when parsing the `static` as well, the `_3()` we get `len(p) == 6`, which is slightly awkward to handle in a function with currently has separate paths for `len(p) > 4` or not.

In the end I put it in a new `p_direct_abstract_declarator_8()`, but I'm happy to revise that if requested.

eliben (approved):

PR #547, opened 2024-07-03
Update c_parser.py to fix string concatenation error

Llewyllen:
Do not remove consecutive double quotes, as it might result in a different string in case of an escaped octal or hexadecimal value at the end of the first string AND a numeric character at the beginning of the second string

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -1915,7 +1915,7 @@ class CParser(PLYParser):
             p[0] = c_ast.Constant(
                 'string', p[1], self._token_coord(p, 1))
         else:
-            p[1].value = p[1].value[:-1] + p[2][1:]
+            p[1].value = p[1].value + p[2]
             p[0] = p[1]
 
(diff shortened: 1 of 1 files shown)


tests/test_c_parser.py
```diff
@@ -1674,7 +1674,7 @@ def test_unified_string_literals(self):
         self.assertEqual(d1, ['Constant', 'string', '"hello"'])
 
         d2 = self.get_decl_init('char* s = "hello" " world";')
-        self.assertEqual(d2, ['Constant', 'string', '"hello world"'])
+        self.assertEqual(d2, ['Constant', 'string', '"hello"" world"'])
```
eliben, line 1677: Unfortunately, this makes no sense in the general case - it goes against language semantics (see the discussion on the issue)

Llewyllen: Sadly the general case is wrong in particular cases and cannot be extended to cover these

PR #548, opened 2024-07-22
patch: fix NoneType error for 'visit_Case' function in c_generator.py…

kmg3821:
… when a case is empty by adding none type checker


eliben (changes requested):
Thanks!

Would it be possible to add a test that triggers this? (fails before the PR, succeeds after)

kmg3821:
**Scenario :** 
C code --> AST --> C code

**Input C code :**
> void foo(int n) {
    switch (n)
    {
    case 0:
    case 1:
        printf("Hello");
        break;
    case 2:
        printf("World");
    default:
        break;
    }
}

**Error occurs when converting AST into C code :**

image

This happens because of the "case 0" statement, which is empty : 

image

**After adding the None type checker :** 
The following C code is successfully generated.
image

kmg3821:
> Thanks!
> 
> Would it be possible to add a test that triggers this? (fails before the PR, succeeds after)

**Scenario :** 
C code --> AST --> C code

**Input C code :**
> void foo(int n) {
    switch (n)
    {
    case 0:
    case 1:
        printf("Hello");
        break;
    case 2:
        printf("World");
    default:
        break;
    }
}

**Error occurs when converting AST into C code :**

image

This happens because of the "case 0" statement, which is empty : 

image

**After adding the None type checker :** 
The following C code is successfully generated.
image

eliben:
Yes, I understand

My question is - can you add a test to https://github.com/eliben/pycparser/blob/main/tests/test_c_generator.py as part of your PR that triggers this behavior and verifies that the change fixes it

kmg3821:
I'm sorry for the confusion, but the problem is not with the c_generator.py. 

The issue lies with the "to_dict" function in c_json.py within the examples folder, which saves None instead of an empty list for the case node type. 

It appears that tests are not performed for the examples, so I did not add the test you requested. 

However, if needed, I can add it.

examples/c_json.py
```diff
@@ -116,6 +116,8 @@ def to_dict(node):
     for child_attr in child_attrs_of(klass):
         if child_attr not in result:
             result[child_attr] = None
+            if result['_nodetype'] == "Case" :
```
eliben, line 119: do an `if...else` here?

eliben (commented):
The PR name should be adjusted also to reflect what it's fixing

@mypalmike can you PTAL?

Issue #528, opened 2024-03-13
Unable to parse labels at end of block


LuciaMartinezGavier:
Raises `pycparser.plyparser.ParseError` when a label is at the end of a block. For example, the following code fails:
```python
from pycparser import CParser

ast =  CParser().parse(
    """
    int main(){
        label:
    }
    """
)
```

As this is valid C, I think it shouldn't throw that exception.

Adding a `;` (or any other code line) after the label fixes this issue
```python
from pycparser import CParser

ast =  CParser().parse(
    """
    int main(){
        label:;
    }
    """
)
```

dgutson:
> Fixed by [#562](https://github.com/eliben/pycparser/pull/562)

any ETA to release?

PR #562, opened 2025-01-08
Support for labels with no statement

ignatirabo:
This pull request is aimed at fixing the issue described here: https://github.com/eliben/pycparser/issues/528

The testing I did seems fine, and I implemented it through two rules to ensure that the representation of `label:` and `label:;` is the same.

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -1581,6 +1581,10 @@ class CParser(PLYParser):
         p[0] = c_ast.Default([p[3]], self._token_coord(p, 1))
 
     def p_labeled_statement_4(self, p):
+        """ labeled_statement : ID COLON SEMI"""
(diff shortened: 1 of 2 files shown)


pycparser/c_parser.py
```diff
@@ -1580,6 +1580,14 @@ def p_labeled_statement_3(self, p):
         """ labeled_statement : DEFAULT COLON pragmacomp_or_statement """
         p[0] = c_ast.Default([p[3]], self._token_coord(p, 1))
 
+    def p_labeled_statement_4(self, p):
+        """ labeled_statement : ID COLON SEMI"""
```
eliben, line 1584: I don't think this rule (_4) should be necessary, since a statement can already be empty. Why do you need it?

eliben (changes requested):
This seems to be introducing a large number of new shift/reduce conflicts for the parser, and it's clear why!

ignatirabo:
Thank you for the all the comments. I imitated the test as you suggested.

Issue #563, opened 2025-02-13
Empty case statements


ignatirabo:
Hello, similarly to https://github.com/eliben/pycparser/issues/528, PyCParser does not accept case statements that do not have statements while GCC does.

Example:
```
int main() {
  int i = 0;
  switch (i) {
    case 0:
      i = 1;
    case 1:
  }
  return i;
}
```

Is it possible for me to do a pull request with this feature?

Thank you in advance.

eliben:
Yes, please go ahead and create a PR

PR #564, opened 2025-02-17
Accept empty 'case' or 'default' labels

ignatirabo:
This PR allows PyCParser to accept programs with `case` and `default` labels with empty statements, as mentioned in https://github.com/eliben/pycparser/issues/563.


eliben (approved):
Thanks!

PR #565, opened 2025-03-05
Respect CPP environmental variable

salahcoronya:
pycparser always defaults to "cpp" as the processor. However, in Gentoo, the symlink from cpp to the compiler is optional (defaults to on though). Before this patch, pycparser would use the cpp implementation specified by caller, if none was passed. it defaulted to "cpp". After this patch, pycparser will use the cpp implementation specified by caller, then whatever CPP environmental variable is set to; if its not set "cpp" will be used as before.

This always sets the test up to use the CPP environmental variable (if set, otherwise, use cpp).

### DIFF
--- a/pycparser/__init__.py
+++ b/pycparser/__init__.py
@@ -11,11 +11,12 @@ __all__ = ['c_lexer', 'c_parser', 'c_ast']
 __version__ = '2.22'
 
 import io
+import os
 from subprocess import check_output
 from .c_parser import CParser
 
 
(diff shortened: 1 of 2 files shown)


eliben (commented):
I'd rather not complicate this code for platform-specific idiosyncrasies. It's extremely simple, and anyone is welcome to write their own `parse_file`.

Note taht you pass `cpp_path` into it, so you can do the `getenv` thing when calling it, if needed

Issue #560, opened 2024-12-04
Incorrect removal of parentheses after the comma operator followed by a compound statement


kousuke3021:
When converting the following code into an Abstract Syntax Tree (AST) and generating code from it using pycparser, the parentheses surrounding a compound statement are removed, resulting in code that fails to compile due to a syntax error.
```
void test(const void *pvd_a_source_data){
  if (0 == pvd_a_source_data) {
    (sizeof (0), ({ if (0) ; else ; }));;
  }
}
```
The generated code appears as follows:
```
void test(const void *pvd_a_source_data){
  if (0 == pvd_a_source_data) {
    (sizeof (0), { if (0) ; else ; });
  }
}
```

eliben:
PRs to fix this are welcome.

Please note that the ability to generate C code back from the AST is not considered a critical capability of pycparser; while I'll accept PRs for it, it's generally lower priority than the parsing part.

PR #566, opened 2025-03-16
fix: preserve parentheses for statement exprs

zawan-ila:
Preserve parentheses around statement expressions when generating C code from AST. Addresses #560


eliben (approved):
Thanks!

Issue #464, opened 2022-05-12
Parsing '//' comments


juanfgarcia:
I'm getting an error when parsing '//' style comments and not with '/*...*/' style

I've been searching and I've found in pycparser/ply/ctokens.py

```
# Comment (C-Style)
def t_COMMENT(t):
    r'/\*(.|\n)*?\*/'
    t.lexer.lineno += t.value.count('\n')
    return t

# Comment (C++-Style)
def t_CPPCOMMENT(t):
    r'//.*\n'
    t.lexer.lineno += 1
    return t
```

But changing t_COMMENT(t) to the // style doesn't work.

Is there any option to compile with the C++ style? If not, do I need to run yacc or other tool to make it work?

eliben:
https://github.com/eliben/pycparser#31interaction-with-the-c-preprocessor

cellularmitosis:
@eliben first, thank you so much for your amazing blog and for pycparser.

> The preprocessor is supposed to remove all comments.

I think I found the root cause behind the high volume of these "comments don't work" issues.

It looks like this is a clang vs gcc issue: gcc's cpp removes comments while clang's does not (and clang is the default for macOS users, which is likely where these issue reports are coming from).

```
$ cat foo.c 
int foo() {
    // this is a comment.
    return 42;
}
```

clang:

```
$ /usr/bin/cpp foo.c 
# 1 "foo.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 423 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "foo.c" 2
int foo() {
    // this is a comment.
    return 42;
}
```

gcc:

```
$ /opt/homebrew/bin/cpp-14 foo.c 
# 0 "foo.c"
# 0 "<built-in>"
# 0 "<command-line>"
# 1 "foo.c"
int foo() {

    return 42;
}
```

This appears to be a bug in clang, as `clang --help` describes the `-C` option, but there appears to be no inverse of this option:

```
  -C                      Include comments in preprocessed output
```

The easy solution for macOS users appears to be:

```
brew install gcc
```

and then

```
parse_file(filename, use_cpp=True, cpp_path='/opt/homebrew/bin/cpp-14')
```

or

```
preprocess_file(filename, use_cpp=True, cpp_path='/opt/homebrew/bin/cpp-14')
```

cellularmitosis:
[...]

PR #569, opened 2025-04-01
Explicit error for comments, with README link

cellularmitosis:
@eliben thanks so much for your blog and for pycparser!

This PR adds an explicit exception for comments, with a hint to the user that clang's cpp does not remove comments, and that simply using gcc's cpp instead will likely solve their problem.

```
$ cat foo.c
int foo() {
    // this is a comment.
    return 42;
}
```

```
$ python3 examples/c_json.py foo.c
pycparser.plyparser.ParseError: foo.c:2:5: Comments are not supported (note: GCC's cpp removes comments, Clang's cpp does not).
```

See also my [comment](https://github.com/eliben/pycparser/issues/464#issuecomment-2768804361) on a related github issue.

Hopefully this will curb the endless flood of "comments don't work" github issues! 😭

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,9 +110,9 @@ Interaction with the C preprocessor
 -----------------------------------
 
 In order to be compilable, C code must be preprocessed by the C preprocessor -
-``cpp``. ``cpp`` handles preprocessing directives like ``#include`` and
+``cpp``. A compatible ``cpp`` handles preprocessing directives like ``#include`` and
 ``#define``, removes comments, and performs other minor tasks that prepare the C
-code for compilation.
+code for compilation. Note that clang's cpp does not remove comments.
 
 For all but the most trivial snippets of C code **pycparser**, like a C
 compiler, must receive preprocessed C code in order to function correctly. If
--- a/README.rst
+++ b/README.rst
@@ -112,7 +112,7 @@ Interaction with the C preprocessor
 In order to be compilable, C code must be preprocessed by the C preprocessor -
 ``cpp``. A compatible ``cpp`` handles preprocessing directives like ``#include`` and
 ``#define``, removes comments, and performs other minor tasks that prepare the C
-code for compilation. Note that clang's cpp does not remove comments.
+code for compilation.
 
 For all but the most trivial snippets of C code **pycparser**, like a C
 compiler, must receive preprocessed C code in order to function correctly. If
--- a/pycparser/c_lexer.py
+++ b/pycparser/c_lexer.py
@@ -209,6 +209,10 @@ class CLexer(object):
 
     bad_octal_constant = '0[0-7]*[89]'
 
+    # comments are not supported
+    unsupported_c89_comment = r'\/\/'
+    unsupported_c99_comment = r'\/\*'
+
     # character constants (K&R2: A.2.5.2)
     # Note: a-zA-Z and '.-~^_!=&;,' are allowed as escape chars to support #line
     # directives with Windows paths as filenames (..\..\dir\file)
@@ -473,6 +477,16 @@ class CLexer(object):
     @TOKEN(bad_octal_constant)
     def t_BAD_CONST_OCT(self, t):
         msg = "Invalid octal constant"
+        self._error(msg, t)
+
+    @TOKEN(unsupported_c89_comment)
+    def t_UNSUPPORTED_C89_COMMENT(self, t):
+        msg = "Comments are not supported (note: GCC's cpp removes comments, Clang's cpp does not)."
+        self._error(msg, t)
+
+    @TOKEN(unsupported_c99_comment)
+    def t_UNSUPPORTED_C99_COMMENT(self, t):
+        msg = "Comments are not supported (note: GCC's cpp removes comments, Clang's cpp does not)."
         self._error(msg, t)
 
     @TOKEN(octal_constant)
--- a/pycparser/c_lexer.py
+++ b/pycparser/c_lexer.py
@@ -210,8 +210,8 @@ class CLexer(object):
     bad_octal_constant = '0[0-7]*[89]'
 
     # comments are not supported
-    unsupported_c89_comment = r'\/\/'
-    unsupported_c99_comment = r'\/\*'
+    unsupported_c_style_comment = r'\/\*'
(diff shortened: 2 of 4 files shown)


pycparser/c_lexer.py
```diff
@@ -209,6 +209,10 @@ def _make_tok_location(self, token):
 
     bad_octal_constant = '0[0-7]*[89]'
 
+    # comments are not supported
+    unsupported_c89_comment = r'\/\/'
```
eliben, line 213: I don't understand the c89/c99 distinction here. Aren't `//` just C++ style comments and `/*` the original C style?

pycparser/c_lexer.py
```diff
@@ -475,6 +479,16 @@ def t_BAD_CONST_OCT(self, t):
         msg = "Invalid octal constant"
         self._error(msg, t)
 
+    @TOKEN(unsupported_c89_comment)
+    def t_UNSUPPORTED_C89_COMMENT(self, t):
+        msg = "Comments are not supported (note: GCC's cpp removes comments, Clang's cpp does not)."
```
eliben, line 484: This error message is way too specific; you have a particular issue you're encountering now, but there are a million other reasons people could try to run pycparser on files with comments. It could point to the repository's README, for example (which has a section on the preprocessor)

eliben (commented):
Thanks. I'm still thinking about this, but here are some initial comments

Also: add tests

pycparser/c_lexer.py
cellularmitosis: I thought it made more sense to describe the comments from a C perspective, rather than a C++ perspective.  I'll update this.

Also, I accidentally swapped c89 / c99.

pycparser/c_lexer.py
cellularmitosis: Thanks for the context, I'll point to the README, but I'd also like to add more context to the README.

Is his diagram an accurate understanding of the situation?  
Screenshot 2025-04-02 at 12 09 23 AM

cellularmitosis:
A bit more context on my findings: invoking clang's preprocessor as `cpp` appears to have different behavior than invoking it as `clang -E`.  In particular, macOS system headers appear to rely on clang extensions, but apparently `cpp` turns them off.

In other words, clang's`/usr/bin/cpp` chokes on any macOS system header, while `clang -E` works (but does not remove comments).

cellularmitosis:
PR updated, tests are still TODO.

cellularmitosis:
Ok, tests added to PR.

This is ready for a re-review.

README.rst
```diff
@@ -110,9 +110,9 @@ Interaction with the C preprocessor
 -----------------------------------
 
 In order to be compilable, C code must be preprocessed by the C preprocessor -
-``cpp``. ``cpp`` handles preprocessing directives like ``#include`` and
+``cpp``. A compatible ``cpp`` handles preprocessing directives like ``#include`` and
 ``#define``, removes comments, and performs other minor tasks that prepare the C
-code for compilation.
+code for compilation. Note that clang's cpp does not remove comments.
```
eliben, line 115: I'm not sure what Clang's cpp is... not `clang -E`, right? Regardless, such a comment will likely go out of date, and I prefer excluding it. The sentence above says enough about what cpp should do

eliben:
> A bit more context on my findings: invoking clang's preprocessor as `cpp` appears to have different behavior than invoking it as `clang -E`. In particular, macOS system headers appear to rely on clang extensions, but apparently `cpp` turns support for them off.
> 
> In other words, clang's`/usr/bin/cpp` chokes on any macOS system header, while `clang -E` works (but does not remove comments).

OK, thanks I understand the context now.

Still, I wouldn't say too much about this in the README. I understand you ran into a frustrating situation, but we shouldn't over-index on it. pycparser has been around for ages now, and this very rarely comes up. The errors you added should improve the situation significantly, but I don't want to maintain additional statements and flowcharts on the README that may go stale at any point

cellularmitosis:
Sorry for the confusion: the flowchart was just for communication during this PR and will not be introduced into the repo.

I picked up on and am trying hard to satisfy the desire to avoid over-indexing on error specificity, but I'm going to be frank: I feel the current PR's user help on this topic is already too minimal.

The current PR:
- adds the phrase "A compatible"
- adds the sentence "Note that clang's cpp does not remove comments."

To communicate my thinking on this topic, here's a possible spectrum of context we could offer the user:

The minimum context the user needs in order to understand (but not fix) the problem is:
- `parse_file()` and `preprocess_file()` require a cpp which strips comments 
- your cpp is not stripping comments

The minimum context to fix the problem would also include:
- gcc is an example of a cpp which strips comments

However, I also feel that the majority of developers are on Macs these days.  A pleasant experience would also include direct, specific advice for that majority:
- `brew install gcc`
- add `cpp_path='/opt/homebrew/bin/cpp-14'` as a parameter to any `parse_file()` or `preprocess_file()` calls

Please understand that the user help in the current PR is already less that what I'd prefer.

cellularmitosis:
Thinking about the staleness issue a bit more, I could change _"Note that clang's cpp does not remove comments"_ to something like _"Gcc includes a cpp which meets these requirements"_.

That would be less likely to go stale, as it is independent of clang's behavior, and I would guess it is very unlikely that gcc's cpp would ever stop stripping comments.

eliben:
Thanks for sharing your thoughts. The PR's largest change is adding an explicit error "comments are not supported" with a README link. A user following this link will read about the requirements of cpp and will notice their cpp doesn't do the job. The text also says one may use `clang -E` instead of `cpp`, and even points to an example.

I really think this is sufficient for now.

By the way, it's entirely possible that Mac's cpp not removing comments is a new thing, because this issue *would* have come up more much frequently otherwise. These things change from time to time, and I want to keep our documentation as resilient as possible to these changes (it already talks quite a bit about the cpp issue!!)

cellularmitosis:
Oh my gosh, I am so sorry, I somehow missed that while clang's `cpp` does _not_ remove comments, `clang -E` _does_ remove comments.

cellularmitosis:
Ok, PR updated to remove _"Note that clang's cpp does not remove comments"_.

eliben (approved):

PR #572, opened 2025-05-03
fix bug with qualifiers on pointers

bbb23exposed:
Fixed a bug where a qualifier on the type being pointed to by a pointer would also be interpreted as a qualifier on that pointer.

Example:

```python
from pycparser import CParser
CParser().parse("const int * x;")
```


eliben (changes requested):
This could really benefit from a couple of tests. Ideally a test would produce the wrong result before this PR and the correct result after it

eliben:
... also note the failing tests. You should run tests locally before sending a PR out

PR #581, opened 2025-09-14
Add support for Python 3.14 and drop EOL 3.8

hugovk:
The Python 3.14 release candidate is out. 🚀 

The 3.14 release manager (👋 hi, that's me!) [said](https://discuss.python.org/t/python-3-14-0rc2-and-3-13-7-are-go/102403):

> ## Call to action
> 
> We **_strongly encourage_** maintainers of third-party Python projects to prepare their projects for 3.14 during this phase, and publish Python 3.14 wheels on PyPI to be ready for the final release of 3.14.0, and to help other projects do their own testing. Any binary wheels built against Python 3.14.0rc1 **_will work_** with future versions of Python 3.14. As always, report any issues to [the Python bug tracker](https://github.com/python/cpython/issues).

Also drop EOL Python 3.8:

https://devguide.python.org/versions/

It was removed from CI in https://github.com/eliben/pycparser/commit/784754462363c6a10bf891d46800fe91d9532b4f, this updates the other metadata.

We could also consider dropping 3.9, it's also EOL in a few weeks.


eliben (commented):
Thanks!

PR #570, opened 2025-04-04
showemptyattrs option on Node.show()

cellularmitosis:
This PR adds a `hideemptyattrs` option to `Node.show()`.  Let me know what you think!

Given `hello.c`:

```c
int main(int argc, char** argv) {
    printf("Hello, world!\n");
    return 0;
}
```

Here's a comparison of all of the options.

default `show()`:

```
ast.show(attrnames=False, hideemptyattrs=False, nodenames=False, showcoord=False):
FileAST: 
  FuncDef: 
    Decl: main, [], [], [], []
      FuncDecl: 
        ParamList: 
          Decl: argc, [], [], [], []
            TypeDecl: argc, [], None
              IdentifierType: ['int']
          Decl: argv, [], [], [], []
            PtrDecl: []
              PtrDecl: []
                TypeDecl: argv, [], None
                  IdentifierType: ['char']
        TypeDecl: main, [], None
          IdentifierType: ['int']
    Compound: 
      FuncCall: 
        ID: printf
        ExprList: 
          Constant: string, "Hello, world!\n"
      Return: 
        Constant: int, 0
```

with `hideemptyattrs`:

```
ast.show(attrnames=False, hideemptyattrs=True, nodenames=False, showcoord=False):
FileAST: 
  FuncDef: 
    Decl: main
      FuncDecl: 
        ParamList: 
          Decl: argc
            TypeDecl: argc
              IdentifierType: ['int']
          Decl: argv
            PtrDecl: 
              PtrDecl: 
                TypeDecl: argv
                  IdentifierType: ['char']
        TypeDecl: main
          IdentifierType: ['int']
    Compound: 
      FuncCall: 
        ID: printf
        ExprList: 
          Constant: string, "Hello, world!\n"
      Return: 
        Constant: int, 0
```

`attrnames=True`:

```
ast.show(attrnames=True, hideemptyattrs=False, nodenames=False, showcoord=False):
FileAST: 
  FuncDef: 
    Decl: name=main, quals=[], align=[], storage=[], funcspec=[]
      FuncDecl: 
        ParamList: 
          Decl: name=argc, quals=[], align=[], storage=[], funcspec=[]
            TypeDecl: declname=argc, quals=[], align=None
              IdentifierType: names=['int']
          Decl: name=argv, quals=[], align=[], storage=[], funcspec=[]
            PtrDecl: quals=[]
              PtrDecl: quals=[]
                TypeDecl: declname=argv, quals=[], align=None
                  IdentifierType: names=['char']
[...]

PR #582, opened 2025-09-15
Update _ast_gen.py to be in sync with c_ast.py

simonlindholm:
#570 updated just c_ast.py.


eliben (approved):
Good catch, thank you!

PR #585, opened 2025-11-25
fix: Parser fails with C23 label syntax

### LINKED ISSUE #576: Parser fails with C23 label syntax.
When trying to parse code that uses C23-specific features, pycparser throws a parse error.
Example:

```
switch() {
    // some code ...

    default:

}
```

but work when changed to:

```
switch() {
    // some code ...

    default:
        ;
}
```

codomposer:
## Summary

Fixes an issue where pycparser was reported to fail on C23-style empty labels such as a bare `default:` inside a `switch`.  
The grammar already supports empty labels; this change adds a focused regression test to ensure that an empty `default:` is parsed correctly and remains supported.

Closes #576 

## Details

- Extend TestCParser_whole_code.test_switch_statement in tests/test_c_parser.py with a new case:

  ```c
  int foo(void) {
      int x = 0;
      switch (x) {
      default:
      }
      return 0;
  }

Contribution by Gittensor, learn more at https://gittensor.io/

### DIFF
--- a/tests/test_c_parser.py
+++ b/tests/test_c_parser.py
@@ -2139,6 +2139,25 @@ class TestCParser_whole_code(TestCParser_base):
 
         self.assertEqual(switch.stmt.block_items, [])
 
+        s4 = r'''
+        int foo(void) {
+            int x = 0;
+            switch (x) {
+            default:
+            }
+            return 0;
+        }
+        '''
+        ps4 = self.parse(s4)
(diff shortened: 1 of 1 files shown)


codomposer:
> Such a test seems to already exist in `test_case_empty_statement`; is it different?

oh, I missed it, 
I just wanted to resolve the issue
I'd like to contribute to the repository, can you please assign another issue which hasn't been resolved yet?

PR #590, opened 2026-01-14
Fix #472: _Alignas does not work with custom types in structs

### LINKED ISSUE #472: _Atomic does not work with custom types
```c
typedef int test_int;

typedef struct
{
  _Alignas(int) _Atomic test_int a1;
  int a2;
} test_struct;
```

Adding the following sequence to c11.c causes the following error:

```
======================================================================
ERROR: test_c11_with_cpp (tests.test_general.TestParsing)
----------------------------------------------------------------------
  File "/Users/user/Developer/pycparser/tests/test_general.py", line 62, in test_c11_with_cpp
    ast = parse_file(self._find_file('c11.c'), use_cpp=True,
  File "/Users/user/Developer/pycparser/pycparser/__init__.py", line 90, in parse_file
    return parser.parse(text, filename)
  File "/Users/user/Developer/pycparser/pycparser/c_parser.py", line 147, in parse
    return self.cparser.parse(
  File "/Users/user/Developer/pycparser/pycparser/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/Users/user/Developer/pycparser/pycparser/ply/yacc.py", line 1199, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/Users/user/Developer/pycparser/pycparser/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/Users/user/Developer/pycparser/pycparser/c_parser.py", line 1931, in p_error
    self._parse_error(
  File "/Users/user/Developer/pycparser/pycparser/plyparser.py", line 67, in _parse_error
    raise ParseError("%s: %s" % (coord, msg))
pycparser.plyparser.ParseError: /Users/user/Developer/pycparser/tests/c_files/c11.c:25:34: before: a1

----------------------------------------------------------------------
Ran 130 tests in 6.860s

FAILED (errors=1)
```

mariam851:
Hi @eliben,

This PR fixes issue #472 where the parser fails when _Alignas is followed by a custom type (typedef) inside a struct.

Key changes:

Updated p_specifier_qualifier_list_6 to use append=True to prevent losing alignment specs.

Added p_specifier_qualifier_list_7 to correctly handle TYPEID (custom types) following an alignment specifier.

All existing tests and the new test case passed successfully.

### DIFF
--- a/pycparser/c_parser.py
+++ b/pycparser/c_parser.py
@@ -961,7 +961,12 @@ class CParser(PLYParser):
     def p_specifier_qualifier_list_6(self, p):
         """ specifier_qualifier_list  : specifier_qualifier_list alignment_specifier
         """
-        p[0] = self._add_declaration_specifier(p[1], p[2], 'alignment')
+        p[0] = self._add_declaration_specifier(p[1], p[2], 'alignment', append=True)
+
+    def p_specifier_qualifier_list_7(self, p):
+        """ specifier_qualifier_list    : specifier_qualifier_list TYPEID
+        """
+        p[0] = self._add_declaration_specifier(p[1], p[2], 'type', append=True)
 
     # TYPEID is allowed here (and in other struct/enum related tag names), because
     # struct/enum tags reside in their own namespace and can be named the same as types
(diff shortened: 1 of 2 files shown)


mariam851:
Hi @eliben,
Thanks for the feedback! I've just updated the PR to address your points:

New Tests: I've added a regression test based on the issue description to tests/test_c_parser.py. I verified it locally and it passes along with the existing test suite. (I'm attaching a screenshot of the test results below).

Screenshot (44)


LALR Conflicts: I checked the parser's report by rebuilding the tables. The count of conflicts is unchanged: it was 1 shift/reduce conflict (dangling-else) before the change, and it remains 1 after the change. No new ambiguities were introduced.

Let me know if everything looks good now!"

PR #599, opened 2026-04-13
Add `py.typed` to designate package as type-aware

Repiteo:
Ensures that the `pycparser` module is properly recognized in type hinting tools by including `py.typed`[^1]. As we're already using `setuptools>=69`, this makes the file automatically included when put in the package directory, so no other changes were necessary

[^1]: https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information


Repiteo:
On second thought, while there *is* a good amount of type-hinting already present, it's not fully integrated just yet. I'll take a stab at implementing the missing type hints first

PR #600, opened 2026-04-19
Implement simple type hints for `_ast_gen.py` and `c_ast.py`

Repiteo:
The typing logic for this was derived exclusively from `_c_ast.cfg`; as such, it's quite rudamentary. Anything more complex would necessitate expanding the configuation with explicit typing data, or making a whitelist for "expected" types. Part of the type validation process meant updating certain semantics or syntax, as well as exposing a few bugs that've since been resolved.

### DIFF
--- a/pycparser/_ast_gen.py
+++ b/pycparser/_ast_gen.py
@@ -97,16 +97,16 @@ class NodeCfg:
         src = f"class {self.name}(Node):\n"
 
         if self.all_entries:
-            args = ", ".join(self.all_entries)
+            args = ", ".join(self._get_typed_arg(entry) for entry in self.all_entries)
             slots = ", ".join(f"'{e}'" for e in self.all_entries)
             slots += ", 'coord', '__weakref__'"
-            arglist = f"(self, {args}, coord=None)"
+            arglist = f"(self, {args}, coord: Coord | None = None)"
         else:
             slots = "'coord', '__weakref__'"
-            arglist = "(self, coord=None)"
+            arglist = "(self, coord: Coord | None = None)"
 
         src += f"    __slots__ = ({slots})\n"
-        src += f"    def __init__{arglist}:\n"
+        src += f"    def __init__{arglist} -> None:\n"
 
         for name in self.all_entries + ["coord"]:
             src += f"        self.{name} = {name}\n"
@@ -114,7 +114,7 @@ class NodeCfg:
         return src
 
     def _gen_children(self):
-        src = "    def children(self):\n"
+        src = "    def children(self) -> tuple[tuple[str, Node], ...]:\n"
 
         if self.all_entries:
             src += "        nodelist = []\n"
@@ -134,7 +134,7 @@ class NodeCfg:
         return src
 
     def _gen_iter(self):
-        src = "    def __iter__(self):\n"
+        src = "    def __iter__(self) -> Generator[Node]:\n"
 
         if self.all_entries:
             for child in self.child:
@@ -158,6 +158,16 @@ class NodeCfg:
         src = "    attr_names = (" + "".join(f"{nm!r}, " for nm in self.attr) + ")"
         return src
 
+    def _get_typed_arg(self, entry):
+        entry_type = (
+            "list[Node]"
+            if entry in self.seq_child
+            else "Node"
+            if entry in self.child
+            else "str"
+        )
+        return f"{entry}: {entry_type} | None"
+
 
 _PROLOGUE_COMMENT = r"""#-----------------------------------------------------------------
 # ** ATTENTION **
@@ -178,9 +188,13 @@ _PROLOGUE_COMMENT = r"""#-------------------------------------------------------
 """
 _PROLOGUE_CODE = r'''
 import sys
-from typing import Any, ClassVar, IO, Optional
+from collections.abc import Callable, Generator
+from typing import Any, ClassVar, IO, TYPE_CHECKING
 
-def _repr(obj):
+if TYPE_CHECKING:
+    from .c_parser import Coord
+
+def _repr(obj: Any | None) -> str:
     """
     Get the representation of an object, with dedicated pprint-like format for lists.
     """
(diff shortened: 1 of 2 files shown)


eliben (commented):
Can you please isolate just the typing changes for the AST so I can review?

No need to add annotations to `ast_gen.py` itself -- it's just a simple script that glues strings together

Repiteo:
Typing changes isolated!

pycparser/c_ast.py
```diff
@@ -63,7 +67,7 @@ def __repr__(self):
 
         return result
 
-    def children(self):
+    def children(self) -> None:
         """A sequence of all children that are Nodes"""
         pass
```
Repiteo, line 72: Typing systems expect subclass return types to match their base class. Given this is an abstract method that's never called, this should be adjusted to:
```suggestion
    def children(self) -> tuple[tuple[str, Node], ...]:
        """A sequence of all children that are Nodes"""
        raise NotImplementedException()
```

pycparser/c_ast.py
```diff
@@ -30,13 +34,13 @@ def _repr(obj):
 
 
 class Node:
-    __slots__ = ()
+    __slots__: ClassVar[tuple[str, ...]] = ()
     """ Abstract base class for AST nodes.
     """
```
Repiteo, line 39: Should be moved below the docstring so it actually can function _as_ a docstring
```suggestion
    """Abstract base class for AST nodes."""
    __slots__: ClassVar[tuple[str, ...]] = ()
```

pycparser/c_ast.py
```diff
@@ -109,7 +113,7 @@ def show(
 
         if self.attr_names:
 
-            def is_empty(v):
+            def is_empty(v: str | None) -> None:
                 v is None or (hasattr(v, "__len__") and len(v) == 0)
```
Repiteo, line 117: This was definitely intended as a conditional, but it never returns
```suggestion
            def is_empty(v: str | None) -> bool:
                return v is None or (hasattr(v, "__len__") and len(v) == 0)
```

pycparser/c_ast.py
```diff
@@ -1326,15 +1451,15 @@ def __iter__(self):
...
 
-    def children(self):
+    def children(self) -> tuple[tuple[str, Node], ...]:
         nodelist = []
         return tuple(nodelist)
 
-    def __iter__(self):
+    def __iter__(self) -> Generator[Node, None, None]:
         return
         yield
```
Repiteo: Not a typing warning/error per se, but this yield is considered unreachable; a way to handle this without making the IDE yell would be
```suggestion
    def __iter__(self) -> Generator[Node, None, None]:
        yield from ()
```

Repiteo (commented):
While the typing changes are now isolated, this consequently re-reveals the bugs that they identified. Here's the most obvious offenders:

pycparser/c_ast.py
```diff
@@ -16,10 +16,14 @@
 
 
 import sys
-from typing import Any, ClassVar, IO, Optional
+from collections.abc import Callable, Generator
+from typing import Any, ClassVar, IO, TYPE_CHECKING
 
+if TYPE_CHECKING:
```
eliben, line 22: Is it necessary to use this check? 
In other parts of pycparser `from typing ....` imports are done unguarded

pycparser/_ast_gen.py
```diff
@@ -178,9 +188,13 @@ def _gen_attr_names(self):
...
 import sys
-from typing import Any, ClassVar, IO, Optional
+from collections.abc import Callable, Generator
+from typing import Any, ClassVar, IO, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .c_parser import Coord
 
-def _repr(obj):
+def _repr(obj: Any | None) -> str:
```
eliben, line 197: What benefit does "Any | None" provide?

pycparser/c_ast.py
```diff
@@ -30,13 +34,13 @@ def _repr(obj):
 
 
 class Node:
-    __slots__ = ()
+    __slots__: ClassVar[tuple[str, ...]] = ()
```
eliben, line 37: is it really necessary to annotate `__slots__`? what check fails if we don't?

pycparser/c_ast.py
```diff
@@ -200,21 +204,27 @@ def generic_visit(self, node: Node):
...
+    def children(self) -> tuple[tuple[str, Node], ...]:
         nodelist = []
         if self.type is not None:
             nodelist.append(("type", self.type))
         if self.dim is not None:
             nodelist.append(("dim", self.dim))
         return tuple(nodelist)
 
-    def __iter__(self):
+    def __iter__(self) -> Generator[Node, None, None]:
```
eliben: Do we really need the `None`s here? aren't they the default?

pycparser/c_ast.py
Repiteo: You _can_ do it unguarded, but it'll cause circular imports down the road. If I tried this without the conditional, `from .c_parser import Coord` would fail for that reason.

Given you're only doing those imports for type hints in pycparser, they should probably be behind a `TYPE_CHECKING` conditional too, but that's outside the scope of this PR

pycparser/_ast_gen.py
Repiteo: Making the typing explicit; that's about it.

pycparser/c_ast.py
Repiteo: You're absolutely right! I completely forgot that it had a smaller option available

pycparser/c_ast.py
Repiteo: mypy gives an error in `__repr__` when trying to iterate over `self.__slots__[:-2]`:
```
Unsupported left operand type for + ("Never") Mypy[operator]
```
Without the hint, it resolves `__slots__` as a tuple with no entries. The hint makes it see a tuple with an indeterminite amount of entries, so it will no longer complain

Repiteo:
Rebased with the `Generator[Node]` fix, but the `ty` checker really didn't like these changes in isolation. This might not be a functional baseline after all, unless you want to actively add more  `None` conditionals/fallbacks

pycparser/_ast_gen.py
eliben: Doesn't `Any` already subsume the `| None` part, making it obsolete?

eliben:
I fixed the missed return you've mentioned, but otherwise I don't feel this is worthwhile at this time. I personally only prefer using `ty` as the typechecker and don't want to be dealing with appeasing multiple type checkers (personal preference, I know it's not for everyone). Can wait a bit more until its functionality improves.

Repiteo:
Fair enough. If I do any future passes, it'll be via `ty` exclusively

PR #606, opened 2026-09-09
Fix empty compound literal (type){} raising ParseError

### LINKED ISSUE #503: Cast an empty compound literal raises ParserError
`(uint8_t[16]){}` (looks like it's C99 compliant statement) raises a `ParserError` (`before: }`), `(uint8_t[16]){0}` doesn't (it's a different logic, but just to test the origin of the issue)

A little bit more context (obfuscated but should be reproducible), just in case:
```
uint8_t test_a() {
  return test_b((uint8_t[16]){});
}

uint8_t test_b(const uint8_t param[16]) {
  return param[0];
}
```

It's difficult for me to debug the C parser in this project, I'd be happy if @eliben or other folks can point me in what direction to dig to fix this, I'd be happy to provide a PR

IMGillusion:
Fix empty compound literal `(type){}` raising ParseError

C99 allows empty initializer braces in compound literals, e.g. `(int){}`.
The compound-literal path in `_parse_postfix_expression` required at least one
initializer item, so `(type){}` failed with `ParseError: Invalid expression`,
while the same empty braces worked in ordinary initializers (`int x[] = {}`).

This mirrors the existing empty-brace handling in `_parse_initializer`.

**Repro (before):**
```python
from pycparser import c_parser
c_parser.CParser().parse('unsigned char f(){return (unsigned char[16]){};}')
# ParseError: Invalid expression
```

**Verification:**
- New regression test `test_empty_compound_literal` — fails on master, passes with the fix
- Full suite: 134 passed, unchanged (2 pre-existing environment failures unrelated to this change)
- Non-empty compound literals and casts are unaffected

Fixes #503


eliben:
Could you rebase your PR on the `main` branch?
I landed a fix that should resolve the CI failures

IMGillusion:
Done - rebased onto main (af3ac872). Full suite is green locally on the new base (137 passed, 14 subtests). Thanks for landing the fix, that should clear the CI failures.

IMGillusion:
CI is green now (full platform matrix). Ready whenever you have a moment to look.

eliben (approved):

